PR #43, opened 2015-01-09
fix typo in method name StaplerRequest.setBindInterceptpr()

darxriggs:


core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -292,10 +292,14 @@ public BindInterceptor getBindInterceptor() {
     }
 
     public BindInterceptor setBindListener(BindInterceptor bindListener) {
-        return setBindInterceptpr(bindListener);
+        return setBindInterceptor(bindListener);
     }
 
     public BindInterceptor setBindInterceptpr(BindInterceptor bindListener) {
```
KostyaSha: @Deprecate it with description.

KostyaSha: javadoc will be inherit, ok.

PR #45, opened 2015-03-05
Add the JSONP Flavor

Grogdunn:
add the JSONP flavour for avoid:
Refused to execute script from 'TARGET URL' because its MIME type ('application/json') is not executable, and strict MIME type checking is enabled

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/export/Flavor.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Flavor.java
@@ -39,6 +39,11 @@ public enum Flavor {
             return new JSONDataWriter(w,config);
         }
     },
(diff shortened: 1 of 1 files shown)


daniel-beck:
Is this really a good idea from a security POV?

Grogdunn:
I don't know if there is another way to execute a JSONP request if the returned type isn't 'application/javascript'.

From security POV if the client don't trust the server is an issue (with or without JSONP) or if the environment is affected by some type of attack or malware.

daniel-beck:
The Jenkins part of this change is covered by the `SecureRequester` protection to JSONP handling, so I guess adding this won't be a problem wrt. CSRF issues.

PR #46, opened 2015-03-12
[FIXED JENKINS-27256] Dispatcher for getX(long) model object methods

tfennelly:
See [JENKINS-27256](https://issues.jenkins-ci.org/browse/JENKINS-27256).

Added dispatcher for `getX(long)` style model methods. Already tested manually but will add a functional test in Jenkins core test harness (yeah, should have done that first etc etc).

@reviewbybees

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/TokenList.java
+++ b/core/src/main/java/org/kohsuke/stapler/TokenList.java
@@ -86,14 +86,14 @@ public final class TokenList {
         return tokens[--idx];
     }
     public int nextAsInt() throws NumberFormatException {
-        long l = nextAsLong();
-        if (l < Integer.MIN_VALUE) {
-            throw new NumberFormatException(String.format("Token '%d' cannot be interpreted as an integer as its value is less than %d.", l, Integer.MIN_VALUE));
-        } else if (l > Integer.MAX_VALUE) {
-            throw new NumberFormatException(String.format("Token '%d' cannot be interpreted as an integer as its value is greater than %d.", l, Integer.MAX_VALUE));
+        long asLongValue = nextAsLong();
+        if (asLongValue < Integer.MIN_VALUE) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/TokenList.java
```diff
@@ -86,12 +86,20 @@ public String prev() {
         return tokens[--idx];
     }
     public int nextAsInt() throws NumberFormatException {
+        long l = nextAsLong();
+        if (l > Integer.MAX_VALUE) {
+            throw new NumberFormatException(String.format("Token '%d' cannot be interpreted as an integer as it's value is greater than %d.", l, Integer.MAX_VALUE));
```
jglick: `its`

core/src/main/java/org/kohsuke/stapler/TokenList.java
```diff
@@ -86,12 +86,20 @@ public String prev() {
         return tokens[--idx];
     }
     public int nextAsInt() throws NumberFormatException {
+        long l = nextAsLong();
+        if (l > Integer.MAX_VALUE) {
```
jglick: What about `l < Integer.MIN_VALUE`?

tfennelly: Good point ... thanks

jglick:
:+1:

jtnord:
:-1: 
"l" (ell) as a variable name shouldn't be used.
return l can be read as return ell or return one and its not obvious which it is and is hard to spot accidental typos.

tfennelly:
Are we happy now James? ffs

jtnord:
:+1: happy chappy

daniel-beck:
https://github.com/stapler/stapler/blob/master/src/site/resources/reference.html#L160 needs to be updated.

PR #47, opened 2015-03-16
Release fixes

jglick:
Trying to make this releasable to OSSRH.

Some temporary POM code pending merge & release of https://github.com/kohsuke/pom/pull/1.

@reviewbybees esp. @kohsuke and @tfennelly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -276,6 +276,7 @@
         <plugins>
           <plugin>
             <artifactId>maven-gpg-plugin</artifactId>
+            <version>1.6</version>
             <executions>
               <execution>
                 <id>sign-artifacts</id>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.kohsuke</groupId>
     <artifactId>pom</artifactId>
-    <version>10</version>
+    <version>12</version>
   </parent>
   <groupId>org.kohsuke.stapler</groupId>
   <artifactId>stapler-parent</artifactId>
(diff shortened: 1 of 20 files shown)


pom.xml
```diff
@@ -171,6 +170,30 @@
         <version>1.2</version>
       </plugin>
     </plugins>
+    <pluginManagement>
+      <plugins>
```
jtnord: there is a site distribution URL so may also need the site plugin and associated bits.

tfennelly:
:+1: 

Curious ... do we actually need `<pre>... </pre>` wrapping `{@code ... }` ? Crazy that javadoc has an issue with `<pre>`.

tfennelly:
oh and ... thanks Jesse !!

pom.xml
jglick: Maybe. I merely corrected the Maven warnings turned up by `clean install`. There might be other warnings displayed during `release:{prepare,perform}` that I did not see yet; AFAIK there is no good way to preview that (one of the reasons `maven-release-plugin` is a dumb design).

jglick:
> do we actually need `<pre>... </pre>` wrapping `{@code ... }` ?

Not positive. Inline Javadoc preview in the NetBeans IDE seemed to require it. I did not bother to check against `javadoc:javadoc` generated HTML. I picked up the idiom [here](http://minds.coremedia.com/2012/05/16/javadoc-hassle-free-embedded-code-using-code/). Anyway it seems harmless.

stephenc:
I have reservations about the level of maven abuse going on here, but given that we want to unblock the release...

:+1:

jglick:
> maven abuse

Alternate suggestions?

stephenc:
>>maven abuse
>
>Alternate suggestions?

Oh it involves reworking custom packaging etc. Nothing quick

PR #48, opened 2015-04-11
[JENKINS-27918] Downgrade stapler-class warning to FINE

jglick:
[JENKINS-27918](https://issues.jenkins-ci.org/browse/JENKINS-27918)

It is confusing and annoying to repeatedly print a warning to the log when Jenkins core still produces `stapler-class` (despite https://github.com/jenkinsci/jenkins/pull/1443) and there is nothing a user or plugin developer can do about it. Was suggested that this be downgraded in https://github.com/stapler/stapler/commit/82b0a45924b083796d04c94661bbef7b80e2d2bf which introduced this warning.

Cf. #39 and https://github.com/jenkinsci/jenkins/pull/1563.

@reviewbybees

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -70,7 +70,7 @@ import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.logging.Logger;
 
-import static java.util.logging.Level.WARNING;
+import static java.util.logging.Level.*;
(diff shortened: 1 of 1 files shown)


daniel-beck:
Should get an issue and be `lts-candidate`.

jglick:
Agreed about the issue (for integration into core). `lts-candidate` seems premature since the next baseline has yet to be selected.

christ66:
:+1:

tfennelly:
:+1:

PR #49, opened 2015-05-02
Use stapler.tmpPath to set upload tmp dir

kinow:
From https://groups.google.com/forum/#!topic/jenkinsci-dev/aNeXlUqS0R8

At the moment, the only way to set the tmp dir for uploads in Stapler is via `java.io.tmpdir`. Apache Commons FileUpload, used by Stapler, lets you define a different directory for that. 

Following the example from MetaClassLoader, this pull request adds another variable used by stapler for that: `stapler.tmpPath`. Tested locally with Jenkins and Stapler in Eclipse workspace. All tests passed.

Thank you

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -862,7 +862,14 @@ public class RequestImpl extends HttpServletRequestWrapper implements StaplerReq
         if(parsedFormData!=null)    return;
 
         parsedFormData = new HashMap<String,FileItem>();
-        ServletFileUpload upload = new ServletFileUpload(new DiskFileItemFactory());
+        DiskFileItemFactory factory = null;
+        String path = System.getProperty("stapler.tmpPath");
+        if (path != null) {
+            factory = new DiskFileItemFactory(DiskFileItemFactory.DEFAULT_SIZE_THRESHOLD, new File(path));
+        } else {
+            factory = new DiskFileItemFactory();
+        }
+        ServletFileUpload upload = new ServletFileUpload(factory);
         try {
             for( FileItem fi : (List<FileItem>)upload.parseRequest(this) )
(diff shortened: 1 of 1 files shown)


daniel-beck:
Seems unnecessary. If you don't want Stapler/Jenkins to use the default temp directory for file uploads (`%TEMP%` on Windows), just point that somewhere else.

kinow:
A good point @daniel-beck. The counterargument could be that a user may want to specify the upload temporary directory without changing the temporary directory for the entire application - though I'm not sure if that use case if very common. 

We could wait for one or two weeks from today and see if anybody will chime in supporting this pull request, otherwise I'm +1 for closing and re-opening later if necessary. How does that sound?

imoutsatsos:
Thank you both @kinow and @daniel-beck. Changing %TEMP% on windows seems to work! The drawback is that you not only need to be an administrator on the Jenkins Windows server, but you also need to mess around with the windows registry if your Jenkins server is running under a separate account (like ours is). It would be nice to have a global configuration option in Jenkins defining the stapler.tmpPath in a more flexible way.

daniel-beck:
> The drawback is that you not only need to be an administrator on the Jenkins Windows server, but you also need to mess around with the windows registry if your Jenkins server is running under a separate account (like ours is).

Any user can define their own environment variables (which override the global defaults) in Control Panel. Log in as that user once, and set the environment.

If the user Jenkins is running as isn't allowed to log in interactively even once, or has other unusual limitations, that's by your choice. Doesn't mean there should be options added just to make life with these artificial limitations easier though.

kohsuke:
In general, I don't buy the argument of "here's another way to do it so this change is unnecessary" especially when there's no real downside in having this somewhat hidden switch.

OTOH, Bruno seemed to have retracted this change himself, and if so I have no reasons to merge it over Daniel's objection.

I'm closing this PR just as a reflection of its current state.

PR #51, opened 2015-06-24
Indicate implementation class in REST API

christ66:
When using the Rest API it is sometimes difficult to know the object that you are working with (Folder, Freestyle project, Maven project, etc.). This PR adds a `class` element which makes knowing the objects class that are being used by the rest api easier.

@reviewbybees @kohsuke

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/export/Model.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Model.java
@@ -190,6 +190,9 @@ public class Model<T> {
                 superBlacklist.add(p.name);
             }
             superModel.writeNestedObjectTo(object, pruner, writer, superBlacklist);
+        } else {
+            writer.name("class");
+            writer.value(object.getClass().getSimpleName());
         }
 
         for (Property p : properties) {
--- a/core/src/main/java/org/kohsuke/stapler/export/Model.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Model.java
@@ -192,7 +192,7 @@ public class Model<T> {
             superModel.writeNestedObjectTo(object, pruner, writer, superBlacklist);
         } else {
             writer.name("$class");
-            writer.value(object.getClass().getSimpleName());
+            writer.value(object.getClass().getName());
         }
 
         for (Property p : properties) {
--- a/core/src/main/java/org/kohsuke/stapler/export/Model.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Model.java
@@ -149,7 +149,7 @@ public class Model<T> {
     /**
      * Writes the property values of the given object to the writer.
      */
-    public void writeTo(T object, DataWriter writer) throws IOException {
+    public void writeTo(@CheckForNull T object, DataWriter writer) throws IOException {
         writeTo(object,0,writer);
     }
 
@@ -159,7 +159,7 @@ public class Model<T> {
      * @param pruner
      *      Controls which portion of the object graph will be sent to the writer.
      */
-    public void writeTo(T object, TreePruner pruner, DataWriter writer) throws IOException {
+    public void writeTo(@CheckForNull T object, TreePruner pruner, DataWriter writer) throws IOException {
         writer.startObject();
         writeNestedObjectTo(object, pruner, writer, Collections.<String>emptySet());
         writer.endObject();
@@ -179,11 +179,11 @@ public class Model<T> {
      *
      * @deprecated as of 1.139
      */
-    public void writeTo(T object, int baseVisibility, DataWriter writer) throws IOException {
+    public void writeTo(@CheckForNull T object, int baseVisibility, DataWriter writer) throws IOException {
         writeTo(object,new ByDepth(1-baseVisibility),writer);
     }
 
-    void writeNestedObjectTo(T object, TreePruner pruner, DataWriter writer, Set<? extends String> blacklist) throws IOException {
+    void writeNestedObjectTo(@CheckForNull T object, TreePruner pruner, DataWriter writer, Set<? extends String> blacklist) throws IOException {
         if (superModel != null) {
             Set<String> superBlacklist = new HashSet<String>(blacklist);
             for (Property p : properties) {
@@ -191,8 +191,10 @@ public class Model<T> {
             }
             superModel.writeNestedObjectTo(object, pruner, writer, superBlacklist);
         } else {
-            writer.name("$class");
-            writer.value(object.getClass().getName());
+            if (object != null) {
+                writer.name("$class");
+                writer.value(object.getClass().getName());
+            }
         }
 
         for (Property p : properties) {
--- a/core/src/main/java/org/kohsuke/stapler/export/Model.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Model.java
@@ -191,7 +191,7 @@ public class Model<T> {
             }
             superModel.writeNestedObjectTo(object, pruner, writer, superBlacklist);
         } else {
-            writer.name("class");
+            writer.name("$class");
             writer.value(object.getClass().getSimpleName());
         }
 
--- a/core/src/main/java/org/kohsuke/stapler/export/SchemaGenerator.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/SchemaGenerator.java
@@ -104,6 +104,7 @@ public class SchemaGenerator {
             cm = ct.complexContent().extension().base(getXmlTypeName(m.superModel.type)).sequence();
         }
 
+        cm.element().name("class").type(XSD.Types.STRING);
         for (Property p : m.getProperties()) {
             Class t = p.getType();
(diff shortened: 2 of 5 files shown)


svanoort:
Question: are there any clients that are coupled to the old schema, which might be broken by adding a new element?  (I am honestly not sure about this -- it's a question I'd ask though.)

jglick:
@svanoort for proper tracking by GitHub, please comment on the PR’s diff tab, not on component commits.

core/src/main/java/org/kohsuke/stapler/export/Model.java
```diff
@@ -190,6 +190,9 @@ void writeNestedObjectTo(T object, TreePruner pruner, DataWriter writer, Set<? e
                 superBlacklist.add(p.name);
             }
             superModel.writeNestedObjectTo(object, pruner, writer, superBlacklist);
+        } else {
+            writer.name("class");
```
jglick: I would suggest `$class` for consistency with the meta-attribute used in form binding.

core/src/main/java/org/kohsuke/stapler/export/Model.java
```diff
@@ -190,6 +190,9 @@ void writeNestedObjectTo(T object, TreePruner pruner, DataWriter writer, Set<? e
...
+            writer.value(object.getClass().getSimpleName());
```
jglick: Better to use the fully-qualified name (`Class.getName`).

core/src/test/java/org/kohsuke/stapler/export/JSONDataWriterTest.java
```diff
@@ -0,0 +1,57 @@
+package org.kohsuke.stapler.export;
+
+import junit.framework.TestCase;
+
+import java.io.IOException;
+import java.io.StringWriter;
+
+public class JSONDataWriterTest extends TestCase {
```
jglick: Use JUnit 4 for new tests please.

core/src/test/java/org/kohsuke/stapler/export/NamedPathPrunerTest.java
```diff
@@ -56,10 +56,10 @@ public void testRange() throws Exception {
...
         Vhew v = new Vhew("view","aaa",jobs);
 
-        assertResult("{jobs:[{name:job0},{name:job1},{name:job2}]}",    v, "jobs[name]{,3}");
-        assertResult("{jobs:[{name:job3},{name:job4},{name:job5}]}",    v, "jobs[name]{3,6}");
-        assertResult("{jobs:[{name:job38}]}",                           v, "jobs[name]{38}");
-        assertResult("{jobs:[{name:job97},{name:job98},{name:job99}]}", v, "jobs[name]{97,}");
+        assertResult("{class:Vhew,jobs:[{class:Jhob,name:job0},{class:Jhob,name:job1},{class:Jhob,name:job2}]}",    v, "jobs[name]{,3}");
+        assertResult("{class:Vhew,jobs:[{class:Jhob,name:job3},{class:Jhob,name:job4},{class:Jhob,name:job5}]}",    v, "jobs[name]{3,6}");
+        assertResult("{class:Vhew,jobs:[{class:Jhob,name:job38}]}",                           v, "jobs[name]{38}");
+        assertResult("{class:Vhew,jobs:[{class:Jhob,name:job97},{class:Jhob,name:job98},{class:Jhob,name:job99}]}", v, "jobs[name]{97,}");
```
jglick: It would be nicer if the meta-attribute could be omitted where it adds no information: where the `@Exported` attribute’s type (or element type) in fact matches the concrete type of the value. In this case for example, that would mean that `class:Jhob` would not be sent, since `Vhew.jobs` is not a heterogenerous list.

Whether the meta-attribute is desirable at top level is a little unclear. Usually it is superfluous, yet there are some cases where you want it. For example, to distinguish different top-level item types, the [XML output](https://jenkins.ci.cloudbees.com/job/core/job/jenkins-core/api/xml) currently gives a hint in the root element name (a mangled version of the class’ `simpleName`, I think), whereas the [JSON output](https://jenkins.ci.cloudbees.com/job/core/job/jenkins-core/api/json) gives no indication. So it is better to include it in this case. Anyway there is only one such attribute per HTTP request—not much effort.

core/src/main/java/org/kohsuke/stapler/export/Model.java
jglick: @svanoort noted a potential NPE here. The `object` parameter of this method should be marked either `@Nonnull` or `@CheckForNull`, according to an inspection of its callers, and the test case should cover null attribute values (and array/collection elements).

stephenc:
:bug: there is a potential NPE, new tests should be JUnit 4 style and there are other comments added by @jglick that need addressing

core/src/main/java/org/kohsuke/stapler/export/Model.java
daniel-beck: @jglick You mean the field that needs to die for some reason?

cniweb:
When will this PR be completed?

jglick:
If and when the author decides to work on it again.

core/src/main/java/org/kohsuke/stapler/export/SchemaGenerator.java
```diff
@@ -104,6 +104,7 @@ private void writeBean(Schema s, Model<?> m) {
             cm = ct.complexContent().extension().base(getXmlTypeName(m.superModel.type)).sequence();
         }
 
+        cm.element().name("class").type(XSD.Types.STRING);
```
jglick: This no longer matches the new name.

core/src/test/java/org/kohsuke/stapler/export/JSONDataWriterTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @ExportedBean public static class X {
+        @Exported public String a = "aval";
+        public String b = "bval";
+        @Exported public String getC() {return "cval";}
+        public String getD() {return "dval";}
+    }
+
+    @Test
+    public void testSimpleUsage() throws Exception {
+        assertEquals("{\"class\":\"X\",\"a\":\"aval\",\"c\":\"cval\"}",
```
jglick: Presumably all of these tests fail now that the attribute name is `$class`? (The CI job [claims to be building PRs](https://jenkins.ci.cloudbees.com/job/core/job/stapler/configure) but apparently is not.)

christ66: As of right now the unit tests are failing. Working on implementing https://github.com/stapler/stapler/pull/51#discussion_r33624844.

jglick:
Actually it seems you cannot use `$class`, as this is not a valid `NameStartChar` for XML output, so that would result in malformed output from `XMLDataWriter`.

stephenc:
> NameStartChar	   ::=   	":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]

So, yes, `$class` would seem to be a :bug: about the only alternative I can think is `_class`... note that as `$` is a valid java name somebody could implement a class with

```
public Object get$class() {
...
}
```

Anyway, so it's not like we didn't risk an over-lap with a plugin's data model... what we want is a low risk of overlap... `_class` or worst case `__class__` would probably be safe enough

christ66:
@stephenc @jglick I can implement with `_class` if everyone approves. I'll fix the unit tests also to reflect this change.

jglick:
Well the original `class` seems fine. (It need not be a Java identifier.) I was striving for consistency with Stapler usage, but since that is not an option, I see no compelling reason for a more exotic-looking attribute.

To avoid extra work while we try to make up our mind, introduce a constant. :-)

kzantow:
FYI - `class` [is a reserved word in ECMA 6](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Reserved_keywords_as_of_ECMAScript_6) and has a lot of overlap with everything in the ecosystem (HTML, Java, etc.). It would definitely be best to avoid it. I'd +1 the use of `_class` for this reason.

jglick:
As noted, using a keyword does not _necessarily_ matter here, since we are just returning XML or JSON…but it might be safer to use something which is a valid Java/JavaScript identifier in case people are doing automatic object bindings or something like that.

kzantow:
Right, or simply using the structure returned, e.g.:

var thisThing = ...
thisThing.class // may fail, would have to use thisThing['class'], which is a pain
thisThing._class // should be fine

kohsuke:
Change subsumed by #61

Issue #52, opened 2015-07-17
getParameter and its family should work correctly with multipart/form-data


kohsuke:
Follow up from https://github.com/jenkinsci/workflow-plugin/pull/150#discussion_r34889145, we should improve these methods so that they work correctly with multipart/form-data

kohsuke:
@tfennelly volunteered to work on this.

kohsuke:
Solved via PR #53

PR #53, opened 2015-07-17
[Fix for issue 52] Added getParameter (and sister function) overrides

tfennelly:
See issue #52.

Updates the getParameter functions to take multipart/form-data form field values into account.

I marked this as a WIP coz we need to add a test or two.


tfennelly:
@kohsuke maybe have a look at this. I added a test.

Issue #32, opened 2014-03-27
@RequirePOST should be mandatory on web methods


jglick:
Since it is common for people to forget to specify `@RequirePOST`, I would suggest that it take a `boolean value() default true;` attribute so that you can use `@RequirePOST(false)` to explicitly indicate that `GET` requests are acceptable on a given web method. (`Processor` would simply ignore the annotation with `!value()`; might require an API addition to `Interceptor` to allow the processor to get the actual annotation.)

Then `RequirePOSTAnnotationProcessor` should check for apparent web methods (whatever `WebMethodAnnotationProcessor` would accept) which lack the annotation, and issue an error, forcing plugin developers to make a decision about each such method.

Alternately, deprecate `@RequirePOST` and make `@WebMethod` mandatory, adding a `boolean requirePOST();` attribute with no default. I tend to prefer this since it makes it much clearer in source code that something is expected to be a web method (and allows tools like IDEs to find all such methods). For binary compatibility, Stapler would need to continue to discover implicit web methods (and handle explicit web methods missing the annotation attribute), but at least the compiler would force them to be marked for newly built plugins. Would need to clarify the status of `getSomething(String)` methods, which `@WebMethod` claims to support yet which do not seem like web methods in the ordinary sense.

kohsuke:
`@WebMethod` is not just for a submission but used for anything where the request handling should be done in Java code. Requiring POST like this is a bad idea IMO.

jglick:
For methods which are intended to be used as `GET` requests, you would write

```java
@WebMethod(name="whatever", requirePost=false)
```

kohsuke:
Now that I've merged #59 I think I captured the intent of this proposal better in #60.

PR #59, opened 2016-02-20
WebMethod improvements

kohsuke:
To help create better REST APIs with POJOs,

* Define a series of HTTP method annotations like `DELETE` and `GET` that lets you write different methods to handle different verbs for the same URL.
* Allow `@WebMethod` on any method, not just those starting with the `do` prefix


kohsuke:
@reviewbybees though I'd like to hear from everyone, not just fellow bees.

core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -94,9 +94,11 @@
     boolean bindAndInvokeAndServeResponse(Object node, RequestImpl req, ResponseImpl rsp, Object... headArgs) throws IllegalAccessException, InvocationTargetException, ServletException, IOException {
         try {
             Object r = bindAndInvoke(node, req, rsp, headArgs);
-            if (getReturnType()!=void.class)
-                renderResponse(req,rsp,node, r);
+            if (getReturnType() != void.class)
+                renderResponse(req, rsp, node, r);
```
recena: Thanks for these changes :+1:

core/src/main/java/org/kohsuke/stapler/FunctionList.java
```diff
@@ -110,7 +120,19 @@ public boolean keep(Function m) {
...
+    }
+
+    /**
+     * Returns {@link Function}s that are either explicitly {@link WebMethod} or
+     * implicitly so (by having its name start with 'do')
+     */
+    public FunctionList webMethods() {
+        return filter(new Filter() {
+            public boolean keep(Function m) {
+                return m.getName().startsWith("do") || m.getAnnotation(WebMethod.class)!=null;
```
recena: `m.getAnnotation(WebMethod.class) != null`

core/src/main/java/org/kohsuke/stapler/MetaClass.java
```diff
@@ -105,8 +105,8 @@
         this.dispatchers.clear();
         ClassDescriptor node = new ClassDescriptor(clazz,null/*TODO:support wrappers*/);
 
-        // check action <obj>.do<token>(...)
-        for( final Function f : node.methods.prefix("do") ) {
+        // check action <obj>.do<token>(...) and other WebMethods
+        for( final Function f : node.methods.webMethods() ) {
```
recena: `for (final ...`

core/src/main/java/org/kohsuke/stapler/MetaClass.java
```diff
@@ -187,26 +187,19 @@ public String toString() {
...
-                    }
-                    public String toString() {
-                        return String.format("%1$s() for url=/%2$s/...",f.getQualifiedName(),name);
-                    }
-                });
-            }
+            dispatchers.add(new NameBasedDispatcher(name) {
+                public boolean doDispatch(RequestImpl req, ResponseImpl rsp, Object node) throws IOException, ServletException, IllegalAccessException, InvocationTargetException {
+                    if(traceable())
+                        traceEval(req,rsp,node,f.getName()+"()");
```
recena:
```
if (traceable()) {
    traceEval(req, rsp, node, f.getName() + "()");
}
```

core/src/main/java/org/kohsuke/stapler/interceptor/Interceptor.java
```diff
@@ -45,6 +46,8 @@ public void setTarget(Function target) {
      * @return
      *      Return value from the method.
      * @throws InvocationTargetException if you want to send e.g. something from {@link HttpResponses}
+     * @throws CancelRequestHandlingException
```
recena: `CancelRequestHandlingException` or `IllegalAccessException`?

core/src/main/java/org/kohsuke/stapler/verb/HttpVerbInterceptor.java
```diff
@@ -0,0 +1,65 @@
...
+            throw new CancelRequestHandlingException();
+    }
+
+    private boolean matches(StaplerRequest request) {
+        String method = request.getMethod();
+
+        for (Annotation a : target.getAnnotations()) {
+            Class<? extends Annotation> t = a.annotationType();
+            InterceptorAnnotation ia = t.getAnnotation(InterceptorAnnotation.class);
+            if (ia !=null && ia.value()==HttpVerbInterceptor.class) {
```
recena: `ia.value() == HttpVerbInterceptor.class`

Really I'm a human, not a coding style bot :panda_face:

recena:
@kohsuke I tried to verify this PR in local, but there are conflicts. Do you plan to solve this?

core/src/main/java/org/kohsuke/stapler/Function.java
kohsuke: Actually unintentional, I got some more IDE settings to tweak, it looks like

core/src/main/java/org/kohsuke/stapler/interceptor/Interceptor.java
kohsuke: `CancelRequestHandlingException` as stated in javadoc. Or maybe I'm missing your point.

core/src/main/java/org/kohsuke/stapler/Function.java
recena: intentional or not, :+1: for your new settings.

jglick:
Somewhere—but how do I find it?—I wrote up an argument that `@WebMethod` should become mandatory, so that we would drop the magical handling of `do*` methods. I had a proposal for how to make this binary compatible, I think using an annotation processor. Also that `@RequirePOST` would become an attribute of `@WebMethod`, or simply the default.

jglick:
Ah! In #32.

core/src/main/java/org/kohsuke/stapler/Function.java
jglick: :-1: on gratuitous reformatting, which makes the diff harder to review.

jglick:
Specifically I advocate that

* you _must_ place `@WebMethod` on any method you expect to be treated in this way—no more surprise exposure of security-sensitive code to the world
  * `name` should be mandatory: a simple Java rename refactoring should not affect behavior
* you _must_ select one or more HTTP verbs—no more forgetting to disallow `GET` on mutators

I think these changes would make for more explicit and safer code. They are not source-compatible but could be made binary-compatible with a little more effort.

core/src/main/java/org/kohsuke/stapler/Function.java
recena: come on @jglick !!! you know perfectly that these little changes improve the readability of the source code.

I know also, you can review this PR even if it has obfuscated :dizzy_face:

jglick: > these little changes improve the readability of the source code

Yes, very slightly, but they significantly _harm_ the readability of the PR. If you want to do a bulk reformat in a separate PR (and introduce a “diff wall” past which merging & cherry-picking becomes impractical), that is another matter.

kohsuke:
I think Jesse is suggesting a change that goes much beyond the scope of this PR.

What this PR does is that it lets us approach that change better than what's stated in issue #32 (namely by having explicit HTTP verb annotations, POST need not be treated any differently from other verbs.)

I'll capture that in a separate ticket for a future consideration, though I don't think I want to make configuration explicit like that.

jglick:
> I think Jesse is suggesting a change that goes much beyond the scope of this PR.

To some extent, yes, but merging this takes the API in a direction from which it might be harder to recover than where we started.

PR #61, opened 2016-02-25
Indicate implementation class in REST API

kohsuke:
This follows up on #51 and bring the work to completion


kohsuke:
@reviewbybees in particular @christ66

kohsuke:
* Renamed to `_class` based on the discussion in #51 
* The behaviour of how/when `_class` property gets written is now user-controllable
* Added a mode that produces `_class` more intelligently

core/src/test/java/org/kohsuke/stapler/export/NamedPathPrunerTest.java
```diff
@@ -56,10 +58,10 @@ public void testRange() throws Exception {
             jobs[i] = new Jhob("job"+i,"aaa","bbb");
         Vhew v = new Vhew("view","aaa",jobs);
 
-        assertResult("{jobs:[{name:job0},{name:job1},{name:job2}]}",    v, "jobs[name]{,3}");
-        assertResult("{jobs:[{name:job3},{name:job4},{name:job5}]}",    v, "jobs[name]{3,6}");
-        assertResult("{jobs:[{name:job38}]}",                           v, "jobs[name]{38}");
-        assertResult("{jobs:[{name:job97},{name:job98},{name:job99}]}", v, "jobs[name]{97,}");
+        assertResult("{_class:Vhew,jobs:[{_class:Jhob,name:job0},{_class:Jhob,name:job1},{_class:Jhob,name:job2}]}",    v, "jobs[name]{,3}");
```
jglick: If the default is `IS_NEEDED` (as I read it), then why is `_class:Jhob` being emitted here, when that is the declared element type? Seems like this mode is not working as advertised.

kohsuke: You caught me. I just updated the tests mechanically without inspecting their outcome. Looks like I got some more work to do here.

cniweb:
And now?

core/src/main/java/org/kohsuke/stapler/export/ClassAttributeBehaviour.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @see DataWriter#startObject(Type, Class)
+     */
+    abstract Class map(Type expected, Class actual);
+
+    String print(Type expected, Class actual) {
+        return print(map(expected,actual));
+    }
+
+    protected String print(Class t) {
```
oleg-nenashev: IMHO makes sense to convert this guy to a nonnull one

core/src/main/java/org/kohsuke/stapler/export/RubyDataWriter.java
```diff
@@ -53,10 +53,10 @@ public void valueNull() throws IOException {
     }
 
     @Override
-    public void startObject() throws IOException {
+    protected void _startObject() throws IOException {
```
daniel-beck: Breaking change?

core/src/main/java/org/kohsuke/stapler/export/JSONDataWriter.java
```diff
@@ -135,7 +138,17 @@ public void endArray() throws IOException {
         close(']');
     }
 
-    public void startObject() throws IOException {
+    public void startObject(Type expected, Class actual) throws IOException {
```
oleg-nenashev: binary compatibility issue

core/src/main/java/org/kohsuke/stapler/export/DataWriter.java
```diff
@@ -53,6 +54,19 @@
     void startArray() throws IOException;
     void endArray() throws IOException;
 
-    void startObject() throws IOException;
```
oleg-nenashev: :bug:  Binary compatibility change in the public interface

core/src/main/java/org/kohsuke/stapler/export/JSONDataWriter.java
```diff
@@ -135,7 +138,17 @@ public void endArray() throws IOException {
...
+        _startObject();
+
+        String t = config.getClassAttribute().print(expected, actual);
+        if (t!=null) {
+            name(TYPE_PROPERTY_NAME);
```
daniel-beck: Inconsistent naming between `getClassAttribute` and `TYPE_PROPERTY_NAME`?

oleg-nenashev:
:bug: since the PR breaks the binary compatibility of the publicly exported API.
* Impact: https://github.com/search?utf8=✓&q=+%22stapler.export.DataWriter%22+%22startObject%28%22&type=Code&ref=searchresults
* Affected project: https://github.com/stapler/stapler-adjunct-nvd3
* Affected project: https://github.com/tatablack/CiInformer

kohsuke:
Hmm, `DataWriter` wasn't meant to be implemented by outside as `Flavor` was supposed to be the contact point, but oh well.

kohsuke:
I will massage the change some more and come back.

core/src/main/java/org/kohsuke/stapler/export/ClassAttributeBehaviour.java
kohsuke: I'll let you fix it up later if this really bothers you.

core/src/main/java/org/kohsuke/stapler/export/RubyDataWriter.java
kohsuke: No it's not, as base class provides the startObject method it needs

kohsuke:
@reviewbybees This change is ready once again

core/src/main/java/org/kohsuke/stapler/export/DataWriter.java
```diff
@@ -53,6 +54,22 @@
...
 
+    /**
+     * Augments the next {@link #startObject()} call by specifying the type information of that object.
+     *
+     * @param expected
+     *      The declared type of the variable that references this object.
+     *      Null if the object is not referenced by anyone, for example when it's the root.
+     * @param actual
+     *      The actual type of the object being written.
+     *      Null if the object is synthetic is has no valid Java type
```
jglick: typo

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -7,7 +7,35 @@
  */
 public class ExportConfig {
     /**
-     * If true, output will be indented to make it easier for humans to understand.
+     * @deprecated
+     *      Use getter and setter
```
jglick: `withPrettyPrint` is not technically a setter, but OK.

core/src/main/java/org/kohsuke/stapler/export/ClassAttributeBehaviour.java
```diff
@@ -0,0 +1,103 @@
...
+ *     <dd>Every object gets the type information explicitly written out
+ *
+ *     <dt>{@link #IF_NEEDED}
+ *     <dd>Type information will be produced only when it is necessary,
+ *         for example when the declared type and the actual type differ.
+ * </dl>
+ *
+ * <p>
+ * In addition, {@link #simple()} can be used to produce just the simple name of the class,
+ * whereas by default {@linkplain Class#getName() full class name} will be printed.
```
jglick: BTW `DescribableHelper` is designed to require/emit a FQN iff there is ambiguity in `simpleName`s amongst implementations of a given extension point. I think that is not easily done here, since Stapler does not have access to extension lists.

core/src/main/java/org/kohsuke/stapler/export/ClassAttributeBehaviour.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Determines if this type attribute should be written.
+     *
+     * @see DataWriter#type(Type, Class)
+     */
+    abstract Class map(Type expected, Class actual);
+
+    String print(Type expected, Class actual) {
```
jglick: :ant: to be clear, please add `@CheckForNull` to each parameter and the return value, since IIUC all can be null.

core/src/main/java/org/kohsuke/stapler/export/DataWriter.java
```diff
@@ -53,6 +54,22 @@
...
+     */
+    void type(Type expected, Class actual) throws IOException;
     void startObject() throws IOException;
```
jglick: Would be clearer to pass `expected` and `actual` to `startObject` itself.

jglick: (deprecating `startObject()` but calling it as the fallback when we get `AbstractMethodError`; in `-source 8` could add a default implementation for `startObject(Type, Class)`)

core/src/test/java/org/kohsuke/stapler/export/ClassAttributeBehaviourTest.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    @Test
+    public void collectionItemWithTypeVariable() throws Exception {
+        String s = write(new CollectionItemWithTypeVariable());
+        assertEquals("{'_class':'CollectionItemWithTypeVariable','foo':[{'x':1,'y':2},{'_class':'Point3D','x':4,'y':5,'z':6}]}",s);
+    }
+
+    @ExportedBean
+    public static class CollectionItemWithTypeVariable<T extends Point> {
+        @Exported public List<T> foo = (List)Arrays.asList(new Point(1,2), new Point3D(4,5,6));
```
jglick: Worth testing a property of type `List<? extends Point>`.

core/src/test/java/org/kohsuke/stapler/export/XMLDataWriterTest.java
```diff
@@ -55,7 +56,7 @@ public void testInheritance() throws Exception {
         @Exported @Override public String generic() {return "sub2";}
     }
     public void testInheritance2() throws Exception { // JENKINS-13336
-        assertEquals("<sub2><basic>super</basic><generic>sub2</generic></sub2>",
+        assertEquals("<sub2 _class='Sub2'><basic>super</basic><generic>sub2</generic></sub2>",
```
jglick: Emitting the lowercased class name as the XML root element name now becomes superfluous, but I guess we cannot compatibly stop and pick a generic root element name like `<root>`.

jglick:
Added some comments and suggestions but :bee: overall.

core/src/main/java/org/kohsuke/stapler/export/ClassAttributeBehaviour.java
kohsuke: We need to have a list of `ExportedBean`s and all its subtypes to be able to make that decision, so it just won't be possible in short of the entire classpath scanning.

bsideup:
@kohsuke will it handle a case of class name conflicts in different packages?

jglick:
> will it handle a case of class name conflicts in different packages?

Yes because the default behavior is `IF_NEEDED` without `.simple()`.

bsideup:
1) I have a class `com.org.my.plugin.MyAction`
2) I serialize it (i.e. for a backup) since it's only one class with name `MyAction` then it will be serialized as `MyAction`
3) then I install another plugin which creates MyAction too, now I have data serialized with packages, right? But old data is unambiguous now

jglick:
Again the default behavior would be to emit something like

```json
"actions": [
  {
    "_class": "com.org.my.plugin.MyAction",
    "someattr": "somevalue"
  }
]
```

Anyway this patch affects the “exported REST” API. It has nothing to do with serialization. In Jenkins that is handled by XStream, which has an unrelated format which always records FQNs where needed.

bsideup:
ok then, just haven't noticed this default behavior in tests, sorry :)

PR #62, opened 2016-03-02
@Exported(merge=true)

kohsuke:
I need the `StaplerFallback`  equivalent for `@Exported` to be able to export two objects together as if it's one object, while properly handling overlaping properties between two objects.

This change adds that via `@Exported(merge=true)`

This change also uses a newly introduced `TreePruner` to implement the blacklisting to reduce cognitive complexity of that fix.

@reviewbybees


scherler:
looks good to me :+1:

PR #64, opened 2016-03-02
Handle method overrides of @WebMethods better

kohsuke:
It is very convenient to be able to define a contract of web-bound object in a base class (often with abstract methods), then allow subtypes to implement those.

We don't want to repeat the HTTP-binding annotations in overriding methods. It's almost like we want to just have Stapler dispatch a call to the definition in the base class (and let `invokevirtual` resolve to the overriding method in subtypes), instead of introspecting a new method and dispatching directly.

To make this work, `ClassDescriptor` is made smarter to recognize methods that are overriding each other, then the annotation inheritance semantics that go beyond JLS is defined so that the overriding method in subtypes behave as if they have all the annotations inherited from the base type method.

See test cases for the examples.

@reviewbybees


core/src/main/java/org/kohsuke/stapler/jsr269/RequirePOSTAnnotationProcessor.java
```diff
@@ -1,107 +0,0 @@
...
-import javax.lang.model.element.Modifier;
-import javax.lang.model.element.TypeElement;
-import javax.lang.model.type.TypeMirror;
-import javax.tools.Diagnostic;
-import org.kohsuke.MetaInfServices;
-import org.kohsuke.stapler.interceptor.RequirePOST;
-
-@SupportedAnnotationTypes("org.kohsuke.stapler.interceptor.RequirePOST")
-@MetaInfServices(Processor.class)
-public class RequirePOSTAnnotationProcessor extends AbstractProcessorImpl {
```
rsandell: I guess this is safe to remove? But a cold chill did run down my spine when I saw it :)

rsandell:
:bee:

core/src/main/java/org/kohsuke/stapler/jsr269/RequirePOSTAnnotationProcessor.java
kohsuke: This enhancement is to make the very usage work that this checker detects & rejects, as illustrated in the test case.

PR #70, opened 2016-03-28
Stapler regression

vivek:
Fixes regressions with XML serialization as reported in https://issues.jenkins-ci.org/browse/JENKINS-33546
* @ExportedBean/@Expoerted annotations not read on Collection classes
* Plural to Singular renaming fails for nested classes

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/export/Property.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Property.java
@@ -184,7 +184,7 @@ public abstract class Property implements Comparable<Property> {
 
         Class c = value.getClass();
 
-        Model model = null;
+        Model model;
         try {
             model = owner.get(c, parent.type, name);
         } catch (NotExportableException ex) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -187,98 +184,97 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
+
+        if(model!=null) {
+            writer.startObject();
             model.writeNestedObjectTo(value, pruner, writer);
-        writer.endObject();
+            writer.endObject();
+        }else if(skipIfFail){
+            writer.startObject();
+            writer.endObject();
+        }
```
kohsuke: Either throw an exception here, or move this whole `if(skipIfFail)` up into the end of the `catch` block

kohsuke:
Memo of the discussion:
* array state handling looks strange.
* test case needs to have all the property names plural to be able to detect when exactly `toSingular` happens

vivek:
Reverted array singularization logic, added tests. All looks good now.

PR #73, opened 2016-04-22
Add StaplerResponseWrapper

kzantow:
... for wrapping functionality downstream if needed (e.g. capturing output by overriding getWriter()).

@reviewbybees esp. @jglick

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
+++ b/core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
@@ -8,6 +8,7 @@ import java.io.Writer;
 import java.net.URL;
 import java.util.Locale;
 
+import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.Cookie;
@@ -36,6 +37,7 @@ public abstract class StaplerResponseWrapper implements StaplerResponse {
     /**
      * Returns the wrapped instance
      */
+    @Nonnull
     public StaplerResponse getWrapped() {
         return wrapped;
     }
--- a/core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
+++ b/core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
@@ -1,0 +1,336 @@
+package org.kohsuke.stapler;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.net.URL;
+import java.util.Locale;
+
+import javax.servlet.ServletException;
+import javax.servlet.ServletOutputStream;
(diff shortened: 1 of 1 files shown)


svanoort:
🐝 - pretty straightforward

core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
```diff
@@ -0,0 +1,336 @@
...
+ * A basic wrapper for a StaplerResponse, e.g. in order to override some method.
+ * This simply delegates all method calls to the wrapped instance.
+ * 
+ * @since TODO
+ */
+@SuppressWarnings("deprecation")
+public abstract class StaplerResponseWrapper implements StaplerResponse {
+    private final StaplerResponse wrapped;
+    
+    public StaplerResponseWrapper(StaplerResponse wrapped) {
```
svanoort: 🐜 but not a bad idea to put a @NotNull to enforce non-nullity of the wrapped response

kzantow: Maybe, but you might need to override `getWrapped()` to get a `StaplerResponse` by some other means (I've had to do this in the `HttpServletResponseWrapper` world before), and having a null `wrapped` member variable is valid in that case.

svanoort: @kzantow Wouldn't it make more sense to just extend StaplerResponse in that case?  You're not really using the wrapper then.

kzantow: You'd still need to forward all the calls to some `StaplerResponse` somewhere. Let's say you wanted to `@Inject StaplerResponse`, you couldn't set the current response and you'd have to provide a wrapper, for example, to look it up from a `ThreadLocal` variable every time.

svanoort: Hmm... okay I can see that case.  On considering again, I think getWrapped() would be the right place to put NotNull, not on initialization.

core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
```diff
@@ -0,0 +1,336 @@
...
+
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.export.Flavor;
+
+import net.sf.json.JsonConfig;
+
+/**
+ * A basic wrapper for a StaplerResponse, e.g. in order to override some method.
+ * This simply delegates all method calls to the wrapped instance.
```
daniel-beck: Wouldn't it make sense to explain how this is supposed to be used?

andresrc:
:bee:

kzantow:
@svanoort 's comment addressed

svanoort:
🐝

oleg-nenashev:
LGTM

PR #74, opened 2016-04-27
BeanInfo doesn't work when there's assymetry in getter vs setter

kohsuke:
In Jenkins this shows up at Slave.nodeProperties whose setter accepts
List but getter returns more specific DescribableList.

@reviewbybees

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -47,12 +47,14 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 import javax.servlet.http.HttpServletResponse;
+import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -815,31 +817,18 @@ private Object bindResolve(Object o, JSONObject src) {
...
-                        continue;
-                    }
 
                     Class<?>[] pt = wm.getParameterTypes();
-
-                    if (pt.length!=1) {
-                        LOGGER.log(WARNING, "Setter method "+wm+" is expected to have 1 parameter");
-                        continue;
-                    }
+                    assert pt.length==1;
```
jtnord: already covered in line 847?

jglick:
Seems dangerous. These things do not work well in `DescribableModel` and it makes it harder to understand what is going on. I would rather just fix the setter to require the more specific type, or add an overload and make that the `@DataBoundSetter`.

In the case of `Slave` there is no such setter. I suppose you mean the `@DataBoundConstructor` for `DumbSlave`. So, IMO, deprecate that overload, introduce a new one requiring `DescribableList`, and make that one the `@DataBoundConstructor`.

jtnord:
:bee: but I generally dislike asserts in production code - especially ones that can not be hit unless you have some memory corruption :-)

kohsuke:
@jglick I did consider that prior to settling on this but it's not that easy. `DescribableList` isn't really injectable because it's not a typical collection type that has a no-arg constructor.

Creating a separate overload for this property requires us to use a different property name, break compatibility, or allow `@DataBoundSetter` to define a property name override, which is less desirable.

I think this is the least path of resistance.

kohsuke:
And I'd also point out that this getter setter asymmetry is already happening. Before the change in https://github.com/jenkinsci/jenkins/issues/2293 the constructor parameter type was already different from that of the getter. So this change is not making things worse in that regard.

kohsuke:
I talked with Jesse offline and he explained to me the context and bigger issue that this touches, which I agreed is a problem that needs to solve. However, I argued that that's more than I'm willing to bite right now and he was OK to look the other way and let me put this in, in order to get other stories moving that hings on this one.

PR #75, opened 2016-06-07
AcceptHeader doesn't honor relative weight in certain cases

vivek:
AcceptHeader.select() gives wrong result if first mime  type is of lower q value than the second parameter mime type.

Given this accept header:

    AcceptHeader ah = new AcceptHeader("    text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8");
    ah.select("text/plain", "text/html");  //gives text/plain, where as it should give text/html as text/html's q is 1 where as text/plain q is 0.8 (*/*)


oleg-nenashev:
LGTM

Issue #76, opened 2016-06-15
Parallel request routing for Blue Ocean


kohsuke:
This issue has its origin in Jenkins Blue Ocean project. In BO, we wanted to expose all `Action`s of a `Build` to the URL space of BO, including its sub-objects, in more REST-ful manner. Such code needs to be able to deal with arbitrary `Action` subtypes that plugins define.

The way we bind those objects to the URL space should mimic the classic UI, thus it should follow regular Stapler routing rules. But BO also need to be able to tweak some aspects of it. For example, `index.jelly` should be replaced by the state transfer of the object. By extension, we also want to disable all other Jelly/Groovy views that are defined on those objects. In the future a need might arise to do some additional tweaking like this.

I thought about various ways of doing it, such as doing this entirely in BO, but I eventually came to believe that such capability should live in Stapler. I think of this capability as a way of wrapping & filtering existing routing rules and making adjustments, and there's really nothing Jenkins specific about it.

This issue captures this task, so that the solution I propose can be discussed in the context of the requirement.

kohsuke:
I'm developing this in #77

kohsuke:
Resolved

PR #77, opened 2016-06-15
[#76] Parallel request routing for Blue Ocean

kohsuke:
Still being developed but this allows comments.

@reviewbybees in particular @vivek


kohsuke:
This change is ready

vivek:
Looks Good 🐝

PR #78, opened 2016-08-24
[JENKINS-37664] Immediately throw an EOF in a boundary error condition

jglick:
[JENKINS-37664](https://issues.jenkins-ci.org/browse/JENKINS-37664)

@reviewbybees

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -149,6 +149,12 @@
       <version>4.5</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>test-annotations</artifactId>
+      <version>1.2</version>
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
🐝 
We may likely need a stable stapler 1.243.x branch since the master 1.244 release is borked.
This fix likely deserves backporting, but I'm not sure 1.244 features are safe enough.

svanoort:
I :bee: lieve this looks reasonable

jglick:
> This fix likely deserves backporting

Not sure. I believe I am now working around it in the _reported_ case in https://github.com/jenkinsci/workflow-job-plugin/pull/18. Of course there could be other cases where this method is called with an illegally large argument, but if these are not being reported as a problem by users then we do not need to backport it aggressively.

PR #79, opened 2016-09-06
allow JSON responses to be compressed when serving with serveStaticResource

daspilker:
To get served with compression, the MIME type of a static resource must either be `TEXT/*` or the file extension must be contained in `TEXT_FILES`. See https://github.com/stapler/stapler/blob/stapler-parent-1.244/core/src/main/java/org/kohsuke/stapler/Stapler.java#L532.

The MIME Type for `.json` files is `application/json`. JSON resources are never compressed since the MIME type is not `text/*` and `json` is not in `TEXT_FILES`.


svanoort:
👍  indeed.

KostyaSha:
@jglick https://groups.google.com/forum/#!topic/jenkinsci-dev/tLGDeaVmxkU

oleg-nenashev:
Post-:bee:

PR #80, opened 2016-10-06
Look for declared fields on superclass

vivek:
Fixes regression where a declared field on a superclass not detected by stapler resulting in 404 when accessed from URL, as reported on https://github.com/jenkinsci/jenkins/pull/2415.


vivek:
@reviewbybees

oleg-nenashev:
Well. @reviewbybees procedure has been violated. But the code is more or less good

PR #81, opened 2016-10-11
[JENKINS-32503] Update dependency of beanutils

KostyaSha:
https://issues.jenkins-ci.org/browse/JENKINS-32503

Testing PR build with dependency update as is.
Jenkins is already on 1.8.3 https://github.com/jenkinsci/jenkins/blob/74c534207ca0b97b0345bf413b63f2450ccc8d04/core/pom.xml#L277-L279
That fails 

---

This change is [Reviewable](https://reviewable.io/reviews/stapler/stapler/81)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
-      <version>1.7.0</version>
+      <version>1.9.3</version>
     </dependency>
     <dependency>
       <groupId>commons-io</groupId>
--- a/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
@@ -441,7 +441,7 @@ public class DataBindingTest extends TestCase {
 
 
     public void testFluentSetter() {
-        FluentSetter fl = bind("{items : 'someItem'}", FluentSetter.class);
+        FluentSetter fl = bind("{items : ['someItem']}", FluentSetter.class);
         assertEquals(Arrays.asList("someItem"), fl.getItems());
     }
 }
--- a/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
@@ -422,7 +422,7 @@ public class DataBindingTest extends TestCase {
         public FluentSetter() {}
 
         /**
-         * WIP This tests passes in stapler, bad fluent doesn't work in jenkins...
+         * WIP This tests passes in stapler, but fluent doesn't work in jenkins...
          * @see FluentPropertyBeanIntrospector
          *
          * It is also possible to transform it to `withXXX()` to have clear vision.
--- a/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
@@ -3,6 +3,8 @@ package org.kohsuke.stapler;
 import junit.framework.TestCase;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.apache.commons.beanutils.FluentPropertyBeanIntrospector;
+import org.junit.Test;
 
 import javax.annotation.PostConstruct;
 import java.lang.reflect.Type;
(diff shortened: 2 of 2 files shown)


KostyaSha:
No PR builds? 0_o

KostyaSha:
https://jenkins.ci.cloudbees.com/job/core/job/stapler/58/
Only one failure :)

KostyaSha:
Commit that introduced that test https://github.com/stapler/stapler/commit/c3004deca59f842614f54203a33bdb2d89ce10d5

KostyaSha:
Any help is welcome ;)

KostyaSha:
This test also fails on 1.8.3 that used in Jenkins.

KostyaSha:
@rsandell maybe you can help? :)

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -398,4 +400,34 @@ public void testDerivedProperty() {
...
+        public FluentSetter() {}
+
+        /**
+         * WIP This tests passes in stapler, but fluent doesn't work in jenkins...
+         * @see FluentPropertyBeanIntrospector
+         *
+         * It is also possible to transform it to `withXXX()` to have clear vision.
+         */
+        @DataBoundSetter
+//        public void setItems(List<String> items) {
```
oleg-nenashev: dead code

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -398,4 +400,34 @@ public void testDerivedProperty() {
...
     }
+
+    public static class FluentSetter {
+        private List<String> items;
+
+        @DataBoundConstructor
+        public FluentSetter() {}
+
+        /**
+         * WIP This tests passes in stapler, but fluent doesn't work in jenkins...
```
oleg-nenashev: CC @vivek

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
KostyaSha: Of course it dead code, because it WIP

jglick:
`DataBindingTest` failing.

KostyaSha:
> DataBindingTest failing.

it's known, i have no power to fix scalarTest and seems i'm missing something...

oleg-nenashev:
@KostyaSha  could you please rebase it to the master

core/pom.xml
```diff
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
-      <version>1.7.0</version>
+      <version>1.9.3</version>
```
jglick: Too new, and see #134.

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -412,4 +414,34 @@ public void testDerivedProperty() {
...
+         * @see FluentPropertyBeanIntrospector
+         *
+         * It is also possible to transform it to `withXXX()` to have clear vision.
+         */
+        @DataBoundSetter
+//        public void setItems(List<String> items) {
+        public FluentSetter setItems(List<String> items) {
```
jglick: Seems potentially useful, but why is it filed here? Should be split into a separate PR.

KostyaSha:
When we tried with oleg it seems passed test. Kohsuke changed databinding
in core, so maybe it already works without beanutils update. Need try
again.

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
KostyaSha: Fluentsetters was exactly the reason to update beanutils that provides fluentbinder used in core... But i had working testcase here and failing in core + unclear test failure stuck me.

core/pom.xml
johnypony3: Can this be revisited? We are running into some serious dependency issues because its at 1.8.3

jglick:
Compare #188.

KostyaSha:
@jglick so if it everything is solved now, then core and test could simplify code with using fluent setters. Requires changes to return `this` instead void.

jglick:
Potentially useful new API, but would need various sorts of work to be mergeable.

PR #83, opened 2016-10-17
Adding a Jenkinsfile

rsandell:


rsandell:
@reviewbybees

rtyler:
declarative Pipeline even, you're so clever :smiley_cat: 


Does this require any special version of the model definition plugin, or will a recent one work fine?

rsandell:
@rtyler This should work with whatever model definition we have installed since I based it on the Jenkinsfile in the PMD plugin and that seems to be running fine  ;)

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
+pipeline {
+    agent label:"docker", docker: "maven:3.3.9-jdk-7"
```
jglick: Prefer to build everything on 8 unless there is a specific reason to use 7.

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
...
+
+    environment {
+      MAVEN_OPTS = "-XX:MaxPermSize=256m -Xmx1024m"
```
jglick: Of course the `MaxPermSize` would be unnecessary on 8. Whether `Mx` is really needed here is TBD. Offhand I would guess that the JRE’s ergonomics are blind to the fact that you are running in a container, and thus the default max heap is going to vary according to host environment.

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    stages {
+        stage("Build") {
+            sh 'mvn clean install -Dmaven.test.failure.ignore=true'
```
jglick: `=true` can be omitted.

jglick: And always pass `-B`.

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
...
+        }
+    }
+
+    postBuild {
+        always {
```
jglick: Even if the build fails? Will this not pick up artifacts from an older build using the same workspace?

jglick: Or cause a spurious and confusing failure in the post-build section after a compilation failure? I would think you would want to do this only if the main build was a success.

Jenkinsfile
rsandell: Done

Jenkinsfile
rsandell: Done

Jenkinsfile
rsandell: Yes, `=true` can be omited but, it [disturbs my circles](https://en.wikipedia.org/wiki/Noli_turbare_circulos_meos!) ;)

Added `-B`

Jenkinsfile
rsandell: Done

stephenc:
:bee:

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
+pipeline {
+    agent label:"docker", docker: "maven:3.3.9-jdk-8"
+
+    environment {
+      MAVEN_OPTS = "-Xmx1024m"
+    }
+
+    stages {
+        stage("Build") {
+            sh 'mvn clean install -B -Dmaven.test.failure.ignore=true'
```
rsandell: Just pure habit

PR #85, opened 2016-11-02
[JENKINS-39414] - Workaround Ruby Runtime Plugin and Jenkins 2.28 compatibility issues

oleg-nenashev:
Stapler 1.266 was formally compatible if we take a set of modules, but we didn't take into account that components like `stapler-jruby` are actually standalone libs being bundled in plugins like Ruby Runtime.

Verifies there is no regression in the Jenkins core itself.

- [x] Add direct unit tests for Field processing for common and Ruby classes
- [x] Suppress error when a plugin uses obsolete lib (e.g. stapler-ruby in Ruby Runtime)

Error suppression is a weird approach, but I cannot offer anything better in the current state. A long-term fix would be to rework Stapler multi-module structure and to somehow ensure that entire stapler code gets delivered via something like core module

https://issues.jenkins-ci.org/browse/JENKINS-39414

@reviewbybees

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
@@ -4,6 +4,7 @@ import org.kohsuke.stapler.Function;
 
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -56,7 +57,13 @@ public final class Klass<C> {
     }
 
     public List<FieldRef> getDeclaredFields() {
-        return navigator.getDeclaredFields(clazz);
+        try {
+            return navigator.getDeclaredFields(clazz);
+        } catch (AbstractMethodError err) {
+            // A plugin uses obsolete version of Stapler-dependent library (e.g. JRuby), which does not offer the method (JENKINS-39414)
+            // TODO: what to do with Logging? The error must be VERY visible, but it will totally pollute system logs
+            return Collections.emptyList();
+        }
     }
 
     /**
--- a/core/src/test/java/org/kohsuke/stapler/lang/KlassTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/lang/KlassTest.java
@@ -1,0 +1,30 @@
+package org.kohsuke.stapler.lang;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.List;
+
+/**
+ * Contains tests for {@link Klass}.
+ * @author Oleg Nenashev.
+ */
+public class KlassTest {
+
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/kohsuke/stapler/lang/Klass.java
```diff
@@ -56,7 +57,13 @@ public Class toJavaClass() {
     }
 
     public List<FieldRef> getDeclaredFields() {
-        return navigator.getDeclaredFields(clazz);
+        try {
+            return navigator.getDeclaredFields(clazz);
+        } catch (AbstractMethodError err) {
```
oleg-nenashev: Actually this handling supresses error I've expected to catch in tests, but tests *on aligned stapler module versions* pass independently of this hack

oleg-nenashev:
Likely https://github.com/jenkinsci/jenkins.rb needs a similar fix

core/src/main/java/org/kohsuke/stapler/lang/Klass.java
```diff
@@ -56,7 +57,13 @@ public Class toJavaClass() {
...
+            // A plugin uses obsolete version of Stapler-dependent library (e.g. JRuby), which does not offer the method (JENKINS-39414)
+            // TODO: what to do with Logging? The error must be VERY visible, but it will totally pollute system logs
```
rsandell: log every 100 occurrence?

rsandell:
:bee:

vivek:
🐝

core/src/main/java/org/kohsuke/stapler/lang/Klass.java
michaelneale: OR just FINE log it?

oleg-nenashev: Both options do not work since we **may** invoke this method dozens and dozens of times during serving a single page. I would rather trigger administrative monitor in Jenkins, but it requires something (e.g. publicly accessible static field with a list of failed navigators). A bit hackish...

oleg-nenashev:
@reviewbybees done

jglick:
:bee:

Actually I think `stapler-jruby` should be a separate repository with its own lifecycle, and we just strive to maintain binary compatibility.

oleg-nenashev:
> Actually I think stapler-jruby should be a separate repository with its own lifecycle, and we just strive to maintain binary compatibility.

Yes, it's a common pitfall of Maven multi-module projects. Actually there are many repositories to be decoupled if we follow such approach.

oleg-nenashev:
@kohsuke @jglick @rsandell @daniel-beck Ping. We need a release ASAP.

rsandell:
I have merged, but only @kohsuke can release as it needs to be signed by his key iirc.

Issue #86, opened 2016-11-07
Add API for getting info about method name and references in KlassNavigator


oleg-nenashev:
Currently one can retrieve list of declared method references from `KlassNavigator`, but there is no way to retrieve names and references to Method implementations. Looks strange.

PR #88, opened 2016-11-07
[JENKINS-39414] - Stapler 1.246 Binary Incompatibility. Part 2

oleg-nenashev:
1.247 didn't fix the issue completely, because there was another incompatible method (`getFunctions()`), which I have missed somehow.

This change intends to solve the binary compat issue:
- [x] - Add default implementations for abstract `KlassNavigator` methods introduced in 1.246
- [x] - Add Javadocs for the involved methods
- [x] - Add tests for `KlassNavigator#getFunctions()`
- [x] - Add tests for `KlassNavigator#getDEclaredMethods()` (generally was not required for this PR)
- [x] - Also provide new API for retrieving info from Methods references #86

I've verified that Ruby Runtime plugin starts correctly and shows UIs without errors in System logs after the fix.

https://issues.jenkins-ci.org/browse/JENKINS-39414

@reviewbybees

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
@@ -86,7 +86,13 @@ public final class Klass<C> {
     }
 
     public List<Function> getFunctions() {
-        return navigator.getFunctions(clazz);
+        try {
+            return navigator.getFunctions(clazz);
+        } catch (AbstractMethodError err) {
+            // A plugin uses obsolete version of Stapler-dependent library (e.g. JRuby), which does not offer the method (JENKINS-39414)
+            // TODO: what to do with Logging? The error must be VERY visible, but it will totally pollute system logs
+            return Collections.emptyList();
+        }
     }
 
     public boolean isArray() {
--- a/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
@@ -88,7 +88,7 @@ public final class Klass<C> {
     }
 
     /**
-     * Gets list of functions declared by the class.
+     * Reports all the methods that can be used for routing requests on this class.
      * @return List of functions. 
      *         May return empty list in the case of obsolete {@link #navigator}, which does not offer the method.
      * @since 1.246
--- a/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/Klass.java
@@ -8,6 +8,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.Nonnull;
 
 /**
  * Abstraction of class-like object, agnostic to languages.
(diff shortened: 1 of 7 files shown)


rsandell:
:bee:

core/src/main/java/org/kohsuke/stapler/lang/Klass.java
```diff
@@ -56,14 +57,15 @@ public Class toJavaClass() {
...
+    @Nonnull
     public List<FieldRef> getDeclaredFields() {
-        try {
-            return navigator.getDeclaredFields(clazz);
-        } catch (AbstractMethodError err) {
-            // A plugin uses obsolete version of Stapler-dependent library (e.g. JRuby), which does not offer the method (JENKINS-39414)
-            // TODO: what to do with Logging? The error must be VERY visible, but it will totally pollute system logs
-            return Collections.emptyList();
-        }
+        return navigator.getDeclaredFields(clazz);
```
scherler: IMHO we need to test this change. You removed the try catch, but are sure that getDeclaredFields never throws an exception?

oleg-nenashev: It was a hack in #85 in order to suppress the binary compatibility issue. Now we solve it by providing default implementations in `KlassNavigator`. Hence this `AbstractMethodError` handler is no longer required

scherler:
@oleg-nenashev ok thanks for the clarification :bee:

oleg-nenashev:
@reviewbybees done

core/src/main/java/org/kohsuke/stapler/lang/KlassNavigator.java
```diff
@@ -78,15 +80,28 @@
...
+    public List<FieldRef> getDeclaredFields(C clazz) {
+        return Collections.emptyList();
+    }
 
     /**
      * Reports all the methods that can be used for routing requests on this class.
+     * @param clazz Class
+     * @return List of the fields functions declared for the class.
+     *         By default this list is empty, {@link KlassNavigator} implementations are responsible to implement it.
+     * @since 1.246
```
jglick: BTW `getDeclaredMethods` was also added as `abstract` somewhat recently (1.220), but I suppose if this caused problems we would have hit them long since.

core/src/main/java/org/kohsuke/stapler/lang/MethodRef.java
```diff
@@ -16,6 +17,18 @@
...
+    
+    /**
+     * Retrieves the referenced method name.
+     * Some implementations (e.g. Ruby) cannot guarantee availability of names for all cases,
+     * so sometimes the name may be missing.
+     * @return Method name. {@code null} if it cannot be determined.
+     * @since 1.248
+     */
+    @CheckForNull
+    public String getName() {
```
jglick: Is there a proposed caller for this, or is it just `KlassTest`?

jruby/src/test/java/org/kohsuke/stapler/jelly/jruby/RubyKlassNavigatorTest.java
```diff
@@ -51,19 +88,66 @@ public void shouldProperlyHandleRubyModules() throws Exception {
...
+            return new RubyFixnum(getRuntime(), 0);
         }
     }
 
-    private static final class MyRubyClass extends RubyClass {
+    private static final class MyRubyClass extends RubyClass {     
         MyRubyClass(Ruby runtime) {
             super(runtime);
-        }
+        }
```
jglick: Gratuitous whitespace changes.

core/src/main/java/org/kohsuke/stapler/lang/KlassNavigator.java
oleg-nenashev: Yes, I decided not to touch that code. Ruby Runtime has a dep on a greater version

PR #90, opened 2016-12-07
Use steps block

michaelneale:
small change for Jenkinsfile


rsandell:
:bee: :+1: and whatever other conventions there are for approving a PR :D

PR #92, opened 2016-12-19
Responses should be exactly as the user intended.

### LINKED ISSUE #91: HttpResponse uses platform dependant line terminators.
[`org.kohsuke.stapler.HttpResponses.html(String)`](https://github.com/stapler/stapler/blob/2faa6d34283a32b371d00b2eefd8862da8eda0ad/core/src/main/java/org/kohsuke/stapler/HttpResponses.java#L208) and  [`org.kohsuke.stapler.HttpResponses.plainText(String)`](https://github.com/stapler/stapler/blob/2faa6d34283a32b371d00b2eefd8862da8eda0ad/core/src/main/java/org/kohsuke/stapler/HttpResponses.java#L224) use platform dependant line terminators in the response, yet the unit test assert that response is using Unix style line ends here and here.

Obviously one is wrong as this causes the unit tests to fail on windows - yet I am not sure which is incorrect, the response or the test.

I am thinking it is more likely to be `HttpResponses` but am not 100% convinced.

```
Failed tests:
  DispatcherTest.testArbitraryWebMethodName:104 expected:<I'm index[]
] but was:<I'm index[
>
  DispatcherTest.testIndexDispatchByName:40 expected:<Hello world[]
] but was:<Hello world[
>
  DispatcherTest.testInheritance:188 expected:<abc[]
] but was:<abc[
>
  DispatcherTest.testInterceptorStage:144 expected:<3,5[]
] but was:<3,5[
>
  DispatcherTest.testInterceptorStageContentTypeWithCharset:161 expected:<3,5[]
] but was:<3,5[
>
  DispatcherTest.testPutInheritance:224 expected:<Hello World![]
] but was:<Hello World![
>
  DispatcherTest.testVerbMatch:67->check:79 expected:<Got GET[]
] but was:<Got GET[
>
```

jtnord:
Stapler was adding a platform dependant line feed to the end of html and plain text
responses.  This was either

1) a bug in the production code
  a) should send as is
  b) should use a non platform specific line end
2) a bug in the unit test that assumed unix line ends

I opted for 1b as the path of least surprise.

Fixes #91 

@reviewbybees

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -7,7 +7,7 @@
   </parent>
   
   <artifactId>stapler</artifactId>
-  <name>Stapler</name>
+  <name>Stapler Core</name>
   <description>Stapler HTTP request handling engine</description>
 
   <build>
--- a/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
+++ b/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
@@ -209,7 +209,9 @@ public class HttpResponses {
         return new HttpResponse() {
             public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
                 rsp.setContentType("text/html;charset=UTF-8");
-                rsp.getWriter().println(literalHtml);
+                PrintWriter pw = rsp.getWriter();
+                pw.print(literalHtml);
+                pw.flush();
             }
         };
     }
@@ -221,7 +223,9 @@ public class HttpResponses {
         return new HttpResponse() {
             public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
                 rsp.setContentType("text/plain;charset=UTF-8");
-                rsp.getWriter().println(plainText);
+                PrintWriter pw = rsp.getWriter();
+                pw.print(plainText);
+                pw.flush();
             }
         };
     }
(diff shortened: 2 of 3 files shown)


jtnord:
test passses for me on both Linux (bash-on-ubuntu-o-windows) and Windows.  Issue does not seem to be with line ends, but just not running the jelly script :-/

jtnord:
> this is going to be incompatible for plugins which were expecting to send text with a newline included

so there are already things broken if they expected that exact behaviour, changing it to use `\n` will not help - they are still broken.

The deprecated mess will leave plugins broken when run on windows until they update - so it really is exactly the same amount of work - the plugin is broken when run on windows and needs to update code either way.  Just one is more work yet noisy, the other is less work and silent (part of the release notes etc that don;t get read)

jglick:
> there are already things broken if they expected that exact behaviour

Not necessarily. Clients can expect a newline not of a particular type.

Still feel uncomfortable with this.

> the plugin is broken when run on windows and needs to update code

Not necessarily. It is up to the author of that plugin to decide what kind of response was intended. Deprecating the current method forces that person to spend a moment thinking about what they want. They may want a response with no newline, or they may want one with NL.

Anyway there was no reported case of any user complaint that I know of. Only test failures, which would be equally well solved by a deprecation + replacement, assuming the test code begins to call the replacement.

jglick:
Took a different approach.

PR #94, opened 2016-12-21
Define an annotation to inject `req.getSubmittedForm()`

kohsuke:
This makes it easier to write a `doConfigSubmit` method that's more amenable for overriding


kohsuke:
@reviewbybees

PR #96, opened 2016-12-29
[Performance] Use invoke dynamic instead of Method.invoke for @Exported

i386:
Use invoke dynamic (`MethodHandles`)  instead of Method.invoke for @Exported. 

`MethodHandles` seem to be what all the cool serialisers like protobuf use for performance reasons. This means that Stapler has to require JDK 7 minimum.

Initial testing seemed to show I could shave 20% off of the time to serve some Blue Ocean rest calls.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ pipeline {
         }
     }
 
-    postBuild {
+    post {
         success {
             archive "**/target/**/*.jar"
             junit '**/target/surefire-reports/*.xml'
--- a/core/src/main/java/org/kohsuke/stapler/MethodHandleCache.java
+++ b/core/src/main/java/org/kohsuke/stapler/MethodHandleCache.java
@@ -1,0 +1,30 @@
+package org.kohsuke.stapler;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.Method;
+import java.util.concurrent.TimeUnit;
+
+public final class MethodHandleCache {
+
+    public static MethodHandle get(Method method) {
+        return METHOD_HANDLES.getUnchecked(method);
+    }
+
+    private static final LoadingCache<Method, MethodHandle> METHOD_HANDLES = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.HOURS).build(new CacheLoader<Method, MethodHandle>() {
+
+        private final Lookup lookup = MethodHandles.lookup();
+
+        @Override
+        public MethodHandle load(Method method) throws Exception {
+            return lookup.unreflect(method);
+        }
+    });
+
+    private MethodHandleCache() {}
+}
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -96,25 +96,25 @@
...
-        </executions>
-        <configuration>
-          <signature>
-            <groupId>org.codehaus.mojo.signature</groupId>
-            <artifactId>java1${java.level}</artifactId>
-            <version>1.0</version>
-          </signature>
-        </configuration>
-      </plugin>
+      <!--<plugin>-->
```
i386: Animal Sniffer complains that the `java.lang.invoke.MethodHandle#invoke` is missing. Not sure what we can do?

i386:
PTAL @vivek

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ pipeline {
         }
     }
 
-    postBuild {
+    post {
```
i386: Unrelated change to make CI pass. Needs https://github.com/stapler/stapler/pull/93 to be merged.

core/src/main/java/org/kohsuke/stapler/MethodHandleCache.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.reflect.Method;
+import java.util.concurrent.TimeUnit;
+
+public final class MethodHandleCache {
+
+    public static MethodHandle get(Method method) {
+        return METHOD_HANDLES.getUnchecked(method);
+    }
+
+    private static final LoadingCache<Method, MethodHandle> METHOD_HANDLES = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.HOURS).build(new CacheLoader<Method, MethodHandle>() {
```
vivek: Maybe configurable number of cache expiry hours, maybe a system property?

vivek:
@i386 It's surprising to see this kind of perf gain just by switch to invokedynamic based invocation. these are pretty basic method invocation (no duck typing involved really, unless we are talking about groovy or jruby like working with duck typing). Or maybe its the cache doing the perf gain as compared to invoke dynamic? IOW did you try comparing perf gain without caching?

i386:
@vivek most of the gains are made when Blue Ocean is using stapler's export package as each @Exported method gets called with reflection. For Blue Ocean, there is a substantial amount of `Method.invoke`'s occurring where the few nanos start to add up compared to using invokedynamic. A good analysis is http://andrewtill.blogspot.com.au/2011/08/using-method-handles.html

vivek:
@i386 Thanks, tested that code locally, indeed MethodHandle speeds things up compared to reflection for large number of iterations. 

Your changes LGTM 🐝

i386:
@vivek did some testing here for 20 mins - it seems the more invocations we make the more the JVM can optimise them.

i386:
Test failure is unrelated and intermittent

kohsuke:
`InstanceFunction` should resolve a method to a handle in the constructor and keep the handle to avoid runtime lookup.

With that change, I don't think caching of `MethodHandle' is necessary, which reduces a complexity and eliminates an arbitrary configuration option (aka cache expiration period.)

I'm happy to make relevant changes and merge. I think it's also fine at this point to require Java7, and hopefully that'll drive some additional source code changes.

i386:
@kohsuke ok that sounds great to me! Appreciate you reviewing these changes 👌

jglick:
> I'm happy to make relevant changes and merge.

The standard procedure is to either request that the original filer do those changes, or to create a new PR subsuming the original one and adding further commits. In either case a fresh round of reviews is required.

PR #97, opened 2017-01-01
Optimization to reduce facet "index" search on every request

i386:
Allows the dispatchers to run first and avoids searching the classpath for index files for each facet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ pipeline {
         }
     }
 
-    postBuild {
+    post {
         success {
             archive "**/target/**/*.jar"
             junit '**/target/surefire-reports/*.xml'
--- a/core/src/main/java/org/kohsuke/stapler/Stapler.java
+++ b/core/src/main/java/org/kohsuke/stapler/Stapler.java
@@ -708,8 +708,6 @@ public class Stapler extends HttpServlet {
             }
         }
 
-        MetaClass metaClass = webApp.getMetaClass(node);
-
         if(!req.tokens.hasMore()) {
             String servletPath = getServletPath(req);
             if(!servletPath.endsWith("/")) {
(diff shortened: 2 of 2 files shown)


i386:
PTAL @vivek

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ pipeline {
         }
     }
 
-    postBuild {
+    post {
```
i386: Depends on https://github.com/stapler/stapler/pull/93

i386:
From what I can tell the test that is failing on CI is intermittent

vivek:
Only concern I have is, if this change will bring slowdown to classic? Anyways, will wait for @kohsuke to comment. I wonder how it relates to https://github.com/stapler/stapler/pull/77, where for example blueocean can pretty much disable jelly path lookup for its rest apis.

i386:
Oh now I know what #77 is all about. Requires a core bump for us though? :(

vivek:
@i386 right, it needs core upgrade. Maybe we can put the code related to PR 77 feature in separate plugin that require upgraded core - this way everything else would continue working, only this parallel routing implementation plugin will need later core version. Something to investigate though.

i386:
If only what was done in #77 was a plugable thing :(

kohsuke:
I haven't look at this change yet, but we need to make sure this change doesn't have a compatibility implications.

i386:
@kohsuke what would be the best way to look at doing this?

kohsuke:
I looked at this change more carefully.

For others, the context of this change is that the current order of lookup in trunk, in case there's no more token in URL path to process, is:

1. Facets get to serve index views (such as `index.jelly/.groovy/.html`) if that's present
2. `doIndex` method gets searchd

... so when routing a request to an object that has no view but `doIndex`, it results in unnecessary search of views before the routing finds the right target.

The proposed code change as of this comment solves this by swapping this order. Even though technically backward incompatible, this reordering is effectively compatible because no valid object would have had index view and `doIndex` method at the same time. This also makes the ordering consistent with `xyz.jelly` vs `doXyz`. The latter wins currently.

What this change fails to accomodate is that the code now pushes index view search so far down the priority list that now it's behind `doDynamic`. This would break a valid user code that defines the index view as well as `doDynamic` as the catch-all.

I think the additional change needed is to fold index view serving into a part of `Dispatcher` like everythng else. The same goes with `HttpDeletable` handling (which probably should be deprecated at this point given `@DELETE` annotation.)

This way, index view can be served at the exact right priority, and code becomes cleaner. In addition, for typical facet implementations like Jelly, the script search can be front-loaded entirely to improve dispatching performance in case a class that doesn't define an index view.

I'll make this follow up change and puts it up as a new PR.

i386:
Thanks mate I look forward to it

kohsuke:
Overtaken by #101

PR #101, opened 2017-01-15
Prefer doIndex over index views

kohsuke:
Adjustment of pull request #97 

The goal of #97 was to improve performance of request routing to objects that do not have index views but do have the `doIndex` method. This is achieved by swapping the priority between those two routes, which was done in 8156216. This is strictly speaking an incompatible change, but in practice acceptable and preferrable because:

1. no valid code could have existed out there that has both index view and the `doIndex` method because the latter would have never been invoked
2. the change makes the relative order of view and web method the same between index (index.jelly vs doIndex) and regular routes (xyz.jelly vs doXyz.)

To make this possible, this PR also converts index view routing and a few other routes that were previously hard-coded in `Stapler.tryInvoke` into `Dispatcher` implementations.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/DirectoryishDispatcher.java
+++ b/core/src/main/java/org/kohsuke/stapler/DirectoryishDispatcher.java
@@ -1,0 +1,42 @@
+package org.kohsuke.stapler;
+
+import javax.servlet.ServletException;
+import java.io.IOException;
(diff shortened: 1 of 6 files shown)


kohsuke:
Tested with the tip of Jenkins to make sure all the tests pass

kohsuke:
@reviewbybees

PR #102, opened 2017-01-15
Fast index view search

kohsuke:
This PR builds on top of PR #101 to improve the request routing performance by resolving the available index view routes upfront.

That is, instead of looking for a presence of `index.jelly` (and the likes) at the request processing time, this change makes that check happen during `MetaClass` construction time.


kohsuke:
Tested with the tip of Jenkins to make sure all the tests pass

kohsuke:
@reviewbybees

PR #103, opened 2017-01-19
Fix use of MethodHandle on web methods which were accidentally static

jglick:
Corrects a regression in #96 caught by tests in https://github.com/jenkinsci/jenkins/pull/2722. [This code](https://github.com/jenkinsci/cvs-plugin/blob/9390d5ae0b07c29c87adc87ae10be302507a5347/src/main/java/hudson/scm/CvsRepository.java#L221) in the CVS plugin is a mistake, but was apparently tolerated before.

@reviewbybees esp. @i386

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Function.java
+++ b/core/src/main/java/org/kohsuke/stapler/Function.java
@@ -35,11 +35,16 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.invoke.MethodHandle;
+import java.lang.invoke.WrongMethodTypeException;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
+import java.util.Arrays;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import static org.kohsuke.stapler.ReflectionUtils.*;
 
@@ -50,6 +55,8 @@ import static org.kohsuke.stapler.ReflectionUtils.*;
  * @author Kohsuke Kawaguchi
  */
 public abstract class Function {
+
+    private static final Logger LOGGER = Logger.getLogger(Function.class.getName());
 
     /**
      * Gets the method name.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -322,15 +329,24 @@ protected MethodHandle handle() {
             return handle;
         }
 
+        @Override
         public Object invoke(StaplerRequest req, StaplerResponse rsp, Object o, Object... args) throws IllegalAccessException, InvocationTargetException {
-            Object[] arguments = new Object[args.length + 1];
-            arguments[0] = o;
-            System.arraycopy(args, 0, arguments, 1, args.length);
+            Object[] arguments;
+            if (Modifier.isStatic(m.getModifiers())) {
```
jglick: Note that `Method.invoke` accepts a non-null first argument for a `static` method (it is just ignored).

core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -322,15 +329,24 @@ protected MethodHandle handle() {
...
+                System.arraycopy(args, 0, arguments, 1, args.length);
+            }
             try {
                 return handle().invokeWithArguments(arguments);
+            } catch (WrongMethodTypeException x) {
+                LOGGER.log(Level.WARNING, handle + " failed on " + o + "." + m + Arrays.toString(arguments), x);
             } catch (Throwable throwable) {
                 throw new InvocationTargetException(throwable);
             }
+            return m.invoke(o, args);
```
jglick: Fallback to previous behavior in case of any other problem.

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -79,6 +79,7 @@ public String getContentType() {
         RequestImpl req = new RequestImpl(new Stapler(), mr, Collections.<AncestorImpl>emptyList(), null);
         new Function.InstanceFunction(getClass().getMethod("doFromStaplerMethod",StaplerRequest.class,int.class,Binder.class))
                 .bindAndInvoke(this,req,null);
+        assertEquals(42, new Function.InstanceFunction(getClass().getMethod("doStaticMethod")).bindAndInvoke(this, req, null));
```
jglick: Yes it is weird to be calling an `InstanceFunction` here, but apparently that is what form validation does. (Trying to use `StaticFunction` here does not work—you get an `ArrayIndexOutOfBoundsException`. Unclear if this is related to unreviewed change by @kohsuke in bd67c1eda4d096c0943a37ff1e84f028bf6348f5.)

core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -322,15 +329,24 @@ protected MethodHandle handle() {
...
+            Object[] arguments;
+            if (Modifier.isStatic(m.getModifiers())) {
+                arguments = args;
+            } else {
+                arguments = new Object[args.length + 1];
+                arguments[0] = o;
+                System.arraycopy(args, 0, arguments, 1, args.length);
+            }
             try {
                 return handle().invokeWithArguments(arguments);
```
jglick: Not clear to me why `handle` is `protected` and not `private`. Unreviewed change by @kohsuke in aca8c023adbc051c80a5e1d785ca76411ff9ed32.

PR #105, opened 2017-02-12
[JENKINS-41729] Allow UTF-8 property files

ikedam:
[JENKINS-41729](https://issues.jenkins-ci.org/browse/JENKINS-41729)

It's difficult to review localization pull requests for Jenkins plugins in github, as property files are encoded.
I have to review them in IDE.

This change makes stapler accept UTF-8 property files for localization, and property files no longer need to be encoded.

Actually, I'm not sure accepting UTF-8 files is the best approach, and I want comments about this change.

### DIFF
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
@@ -28,8 +28,10 @@ import org.kohsuke.stapler.WebApp;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.StringWriter;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


ikedam:
I found I had to find the way to control the negotiating language.
I'll reopen this after I find that way.

ikedam:
I updated the test (I needed to set `LocaleProvider`) and reopened the request.

ikedam:
@jglick You're right. I added a test for a case using properties file encoded with ISO-8859-1, and found that it didn't work.
Worse, I found that Java doesn't detect it fails to decode texts (it doesn't throw exceptions!) and I cannot rely on falling-back codes like:
```
try {
  readWithISO8859_1();
} catch(FailedToDecodeException e) { // This (or something like this) doesn't happen!
  readWithUtf8();
}
```

Rather, I should use [XML property files]( https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#loadFromXML-java.io.InputStream- ) instead of *.properties and I believe it resolves encoding problems as XML files are declared with their encodings (`<?xml encoding="utf-8"?>`).

ikedam:
Replaced with #171

PR #106, opened 2017-02-13
JENKINS-40088# Configurable stapler serialization behavior

vivek:
This PR bring customization feature where plugins like blueocean can configure stapler and participate during property serialization and decide whether to fail or skip serialization. 

Implementation is pretty much based on @i386 [proposal](https://docs.google.com/document/d/1AalzfYZ0HDfwtFMUNaw_okWMLaaQ0S5Q1o6L4taevL4/edit#heading=h.vr11cprbtd0) sent at dev list some time back. Further this custom serialization approach is followup of [PR 89](https://github.com/stapler/stapler/pull/89#issuecomment-272307140) to fix it in generic way.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
@@ -23,26 +23,6 @@
 
 package org.kohsuke.stapler;
 
-import com.jcraft.jzlib.GZIPOutputStream;
-import net.sf.json.JsonConfig;
-import org.apache.commons.io.IOUtils;
-import org.kohsuke.stapler.compression.CompressionFilter;
-import org.kohsuke.stapler.compression.FilterServletOutputStream;
-import org.kohsuke.stapler.export.DataWriter;
-import org.kohsuke.stapler.export.ExportConfig;
-import org.kohsuke.stapler.export.Flavor;
-import org.kohsuke.stapler.export.Model;
-import org.kohsuke.stapler.export.ModelBuilder;
-import org.kohsuke.stapler.export.NamedPathPruner;
-import org.kohsuke.stapler.export.TreePruner;
-import org.kohsuke.stapler.export.TreePruner.ByDepth;
-
-import javax.annotation.Nonnull;
-import javax.servlet.ServletException;
-import javax.servlet.ServletOutputStream;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import javax.servlet.http.HttpServletResponseWrapper;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -56,6 +36,26 @@ import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import javax.annotation.Nonnull;
+import javax.servlet.ServletException;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
+
+import com.jcraft.jzlib.GZIPOutputStream;
+import net.sf.json.JsonConfig;
+import org.apache.commons.io.IOUtils;
+import org.kohsuke.stapler.compression.CompressionFilter;
+import org.kohsuke.stapler.compression.FilterServletOutputStream;
+import org.kohsuke.stapler.export.DataWriter;
+import org.kohsuke.stapler.export.ExportConfig;
+import org.kohsuke.stapler.export.Flavor;
+import org.kohsuke.stapler.export.Model;
+import org.kohsuke.stapler.export.ModelBuilder;
+import org.kohsuke.stapler.export.NamedPathPruner;
+import org.kohsuke.stapler.export.TreePruner;
+import org.kohsuke.stapler.export.TreePruner.ByDepth;
 
 /**
  * {@link StaplerResponse} implementation.
@@ -231,7 +231,7 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
 
     @SuppressWarnings({"unchecked", "rawtypes"}) // API design flaw prevents this from type-checking
     public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flavor) throws ServletException, IOException {
-        serveExposedBean(req, exposedBean, new ExportConfig.Builder(flavor).prettyPrint(req.hasParameter("pretty")).build());
+        serveExposedBean(req, exposedBean, new ExportConfig().withFlavor(flavor).withPrettyPrint(req.hasParameter("pretty")));
     }
 
     @Override
--- a/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
@@ -231,12 +231,13 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
 
     @SuppressWarnings({"unchecked", "rawtypes"}) // API design flaw prevents this from type-checking
     public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flavor) throws ServletException, IOException {
-        serveExposedBean(req, exposedBean, flavor, new ExportConfig());
+        serveExposedBean(req, exposedBean, new ExportConfig.Builder(flavor).prettyPrint(req.hasParameter("pretty")).build());
     }
 
     @Override
-    public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flavor, ExportConfig config) throws ServletException, IOException {
+    public void serveExposedBean(StaplerRequest req, Object exposedBean, ExportConfig config) throws ServletException, IOException {
         String pad=null;
+        Flavor flavor = config.getFlavor();
         setContentType(flavor.contentType);
         Writer w = getCompressedWriter(req);
 
@@ -265,9 +266,6 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
             }
             pruner = new ByDepth(1 - depth);
         }
-
-        config.prettyPrint = req.hasParameter("pretty");
-
         DataWriter dw = flavor.createDataWriter(exposedBean, w, config);
         if (exposedBean instanceof Object[]) {
             // TODO: extend the contract of DataWriter to capture this
@@ -282,7 +280,6 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
 
         if(pad!=null) w.write(')');
         w.close();
-
     }
 
     private void writeOne(TreePruner pruner, DataWriter dw, Object item) throws IOException {
--- a/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
@@ -23,39 +23,39 @@
 
 package org.kohsuke.stapler;
 
+import com.jcraft.jzlib.GZIPOutputStream;
 import net.sf.json.JsonConfig;
+import org.apache.commons.io.IOUtils;
 import org.kohsuke.stapler.compression.CompressionFilter;
 import org.kohsuke.stapler.compression.FilterServletOutputStream;
 import org.kohsuke.stapler.export.DataWriter;
 import org.kohsuke.stapler.export.ExportConfig;
-import org.kohsuke.stapler.export.NamedPathPruner;
 import org.kohsuke.stapler.export.Flavor;
 import org.kohsuke.stapler.export.Model;
 import org.kohsuke.stapler.export.ModelBuilder;
-import org.apache.commons.io.IOUtils;
+import org.kohsuke.stapler.export.NamedPathPruner;
+import org.kohsuke.stapler.export.TreePruner;
+import org.kohsuke.stapler.export.TreePruner.ByDepth;
 
 import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpServletResponseWrapper;
-import javax.servlet.http.HttpServletRequest;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.OutputStreamWriter;
 import java.io.InputStream;
-import java.io.PrintWriter;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
 import java.io.Writer;
-import java.net.URL;
 import java.net.HttpURLConnection;
-import com.jcraft.jzlib.GZIPOutputStream;
+import java.net.URL;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import org.kohsuke.stapler.export.TreePruner;
-import org.kohsuke.stapler.export.TreePruner.ByDepth;
 
 /**
  * {@link StaplerResponse} implementation.
@@ -231,6 +231,11 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
 
     @SuppressWarnings({"unchecked", "rawtypes"}) // API design flaw prevents this from type-checking
     public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flavor) throws ServletException, IOException {
+        serveExposedBean(req, exposedBean, flavor, new ExportConfig());
+    }
+
+    @Override
+    public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flavor, ExportConfig config) throws ServletException, IOException {
         String pad=null;
         setContentType(flavor.contentType);
         Writer w = getCompressedWriter(req);
@@ -239,7 +244,7 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
             pad = req.getParameter("jsonp");
             if(pad!=null) w.write(pad+'(');
         }
-        
+
         TreePruner pruner;
         String tree = req.getParameter("tree");
         if (tree != null) {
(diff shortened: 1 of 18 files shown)


i386:
@jglick PTAL :)

jglick:
I suppose you meant this to be @reviewbybees.

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
```diff
@@ -239,7 +244,7 @@ public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flav
             pad = req.getParameter("jsonp");
             if(pad!=null) w.write(pad+'(');
         }
-        
+
```
jglick: please avoid gratuitous diff hunks

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
```diff
@@ -278,6 +282,7 @@ public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flav
 
         if(pad!=null) w.write(')');
         w.close();
+
```
jglick: ditto

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
```diff
@@ -261,7 +266,6 @@ public void serveExposedBean(StaplerRequest req, Object exposedBean, Flavor flav
             pruner = new ByDepth(1 - depth);
         }
 
-        ExportConfig config = new ExportConfig();
         config.prettyPrint = req.hasParameter("pretty");
```
jglick: Smells wrong to be mutating the config that was passed in to this method.

core/src/main/java/org/kohsuke/stapler/StaplerResponse.java
```diff
@@ -173,10 +175,30 @@
...
+     * This method performs the complete output from the header to the response body.
+     * If the flavor is JSON, this method also supports JSONP via the {@code jsonp} query parameter.
+     *
+     * <p>The {@code depth} parameter may be used to specify a recursion depth
+     * as in {@link Model#writeTo(Object,int,DataWriter)}
+     *
+     * <p>As of 1.146, the {@code tree} parameter may be used to control the output
+     * in detail; see {@link NamedPathPruner#NamedPathPruner(String)} for details.
+     *
+     * <p> {@link ExportConfig} is passed by the caller to control serialization behavior
```
jglick: I would think it more intuitive for `Flavor` to be a field of the `ExportConfig`.

core/src/main/java/org/kohsuke/stapler/export/DataWriter.java
```diff
@@ -69,6 +70,8 @@
     void startObject() throws IOException;
     void endObject() throws IOException;
 
+    @Nonnull ExportConfig getExportConfig();
```
jglick: Not binary-compatible but I suppose we assume there are no external implementations?

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -38,4 +45,22 @@ public ExportConfig withClassAttribute(ClassAttributeBehaviour cab) {
         this.classAttribute = cab;
         return this;
     }
+
+    public ExportInterceptor getExportInterceptor() {
+        return exportInterceptor;
+    }
+
+    public ExportConfig withExportInterceptor(ExportInterceptor interceptor){
```
jglick: Missing Javadoc.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -38,4 +45,22 @@ public ExportConfig withClassAttribute(ClassAttributeBehaviour cab) {
...
+        this.exportInterceptor = interceptor;
+        return this;
+    }
+
+    public ExportConfig withSkipIfFail(boolean skipIfFail){
```
jglick: Ditto.

core/src/main/java/org/kohsuke/stapler/export/ExportInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Allows caller to intercept exporting of properties
+ *
+ * @author Vivek Pandey
+ * @author James Dumay
+ */
+public abstract class ExportInterceptor {
+    /**
+     * Subclasses must call {@link Property#getValue(Object)}  to retrieve the property
```
jglick: missing `.` to terminate summary sentence

core/src/main/java/org/kohsuke/stapler/export/Model.java
```diff
@@ -37,12 +42,6 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import javax.annotation.CheckForNull;
-import javax.annotation.Nullable;
-
-import com.google.common.base.Predicate;
-import org.kohsuke.stapler.export.TreePruner.ByDepth;
-
```
jglick: revert

core/src/main/java/org/kohsuke/stapler/export/ExportInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public abstract Object getValue(Property property, Object model) throws IOException;
+
+    public static final ExportInterceptor DEFAULT = new ExportInterceptor() {
+        @Override
+        public Object getValue(Property property, Object model) throws IOException {
+            try {
+                return property.getValue(model);
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                throw  new IOException("Failed to write " + property.name, e);
```
jglick: generally you want

```java
throw new IOException("Failed to write " + property.name + ": " + e, e);
```

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -147,18 +156,11 @@ public void writeTo(Object object, TreePruner pruner, DataWriter writer) throws
...
-            IOException x = new IOException("Failed to write " + name);
-            x.initCause(e);
-            throw x;
-        } catch (InvocationTargetException e) {
-            IOException x = new IOException("Failed to write " + name);
-            x.initCause(e);
-            throw x;
-        }
+    /**
+     *  Ignores property if {@link ExportInterceptor#getValue(Property, Object)} throws NotExportableException.
```
jglick: Huh? It does not seem to do anything of the sort. In fact why does this method exist? Inline it!

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -14,6 +14,13 @@
 
     private ClassAttributeBehaviour classAttribute = ClassAttributeBehaviour.IF_NEEDED;
 
+    /**
+     * Interceptor used to check if given model object or it's property should be included in serialization
+     */
+    private ExportInterceptor exportInterceptor = ExportInterceptor.DEFAULT;
+
+    private boolean skipIfFail = false;
```
jglick: Seems like this should be another method of `ExportInterceptor`, called I suppose when supplied with something which is not `@ExportedBean` or `CustomExportedBean`. Or at a higher level I would expect the whole handling of a bean to go through `ExportInterceptor`. IOW this looks like the wrong level of abstraction.

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
vivek: ack

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
vivek: yes, fixing this.

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
vivek: ack.

core/src/main/java/org/kohsuke/stapler/StaplerResponse.java
vivek: +1

core/src/main/java/org/kohsuke/stapler/export/DataWriter.java
vivek: Thats right. running this [query](https://github.com/search?utf8=✓&q=org%3Ajenkinsci+org.kohsuke.stapler.export.DataWriter+in%3Afile+&type=Code&ref=searchresults) on GitHub tells no external implemention - yet.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
vivek: ack.

core/src/main/java/org/kohsuke/stapler/export/ExportInterceptor.java
vivek: ack

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
vivek: ack.

core/src/main/java/org/kohsuke/stapler/export/ExportInterceptor.java
vivek: I copied it from somewhere in stapler but happy to make this change.

core/src/main/java/org/kohsuke/stapler/export/Property.java
vivek: Removed that method and comment and inlined it.

core/src/main/java/org/kohsuke/stapler/export/Model.java
vivek: ack

vivek:
@jglick All your comments are addressed, PTAL.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -1,66 +1,146 @@
...
+         *
+         * @return {@link Builder} instance
+         */
+        public Builder prettyPrint(boolean prettyPrint){
+            this.prettyPrint = prettyPrint;
+            return this;
+        }
+
+        /**
+         * If true serialization error will be ignored
```
jglick: I am still a bit uneasy about this: it feels like rather than a hard-coded flag, this should just be an aspect of the behavior of `ExportInterceptor`.

Now currently `ExportInterceptor` is only asked to `getValue` on a single attribute, which is too late to defend against, say, an `Object` which is neither an `@ExportedBean` nor a `CustomExportedBean`. And the handling of those cases is hard-coded in Stapler core and not customizable—except via this flag.

So perhaps that means that the design of `ExportInterceptor` is wrong—too narrowly conceived. Rather than being asked to handle individual attributes, it should be asked to deal with the entire “bean”. The default implementation would check for `CustomExportedBean` and do a replacement if indicated; then fail if not `@ExportedBean`; then produce a `Model`, iterate its properties, and tell the `DataWriter` to create an object with all those properties. A custom implementation could be more or less lenient about various error conditions, or do something else entirely when encountering certain bean types (perhaps delegating to the default interceptor in other cases).

@kohsuke may want to weigh in here.

jglick:
Also ideally you should have a `blueocean-plugin` PR downstream of this demonstrating how you intend to consume a proposed API to fix a particular behavioral problem.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
vivek: >So perhaps that means that the design of ExportInterceptor is wrong—too narrowly conceived.

I would say its more fine grained. We need this to control how serialization of properties are handled than coarse grained bean serialization. Implementations can pretty much control whether object to be serialized is CustomExportedBean or is not ExportedBean etc. 

Maybe we can enhance`ExportInterceptor` to handle top level bean serialization but it can come later on, doesn't have to be this PR.

vivek:
@jglick I have changes on a branch, will incorporate latest changes and open a PR.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
jglick: > drop this flag?

You are not using it anyway?

> Maybe we can enhance `ExportInterceptor` to handle top level bean serialization but it can come later on

Perhaps so.

Would still like a look from someone else, ideally @kohsuke.

vivek: > You are not using it anyway?
Well, I put it there thinking it might be useful for us in case we want to ignore errors (NotExportableException kind), since its something we probably don't want to expose as mentioned in this PR, we can take care of it in BO implementation of `ExportInterceptor`.

vivek:
@jglick corresponding Blueocean PR https://github.com/jenkinsci/blueocean-plugin/pull/853.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -1,41 +1,146 @@
...
+        this.prettyPrint = prettyPrint;
+        this.classAttribute = (classAttribute == null) ? ClassAttributeBehaviour.IF_NEEDED.simple(): classAttribute;
+        this.exportInterceptor = exportInterceptor == null ? ExportInterceptor.DEFAULT :exportInterceptor;
+        this.skipIfFail = skipIfFail;
     }
 
     /**
      * If true, output will be indented to make it easier for humans to understand.
      */
-    public ExportConfig withPrettyPrint(boolean prettyPrint) {
```
oleg-nenashev: Has it been removed at all? Compatibility issue then

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -1,41 +1,146 @@
...
+     * Controls the behaviour of the class attribute to be produced.
+     */
     public ClassAttributeBehaviour getClassAttribute() {
         return classAttribute;
     }
 
     /**
      * Controls the behaviour of the class attribute to be produced.
      */
-    public ExportConfig withClassAttribute(ClassAttributeBehaviour cab) {
```
oleg-nenashev: Same, seems to be compat issue

vivek: @oleg-nenashev ok, yeah its exposed in `Flavor.createDataWriter()` public methods and can be used from outside.  I missed that, will revert it as I don't want to conflate this PR and keep it focused on what its trying to solve. cc: @jglick.

vivek:
@oleg-nenashev ExportConfig compatibility related concerns have been addressed. I have reverted to how it was with enhancements related to this PR. PTAL.

kohsuke:
I talked to @vivek at length and here's how I position this change.

When I think about how the error handling during the serialization should be customizable, it should come in a spectrum:
- on the low end, we should have a boolean flag that makes "the whole thing as lenient and failure tolerant as possible."
- on the high end, we should have a custom serializer that allows the serialization of an object to be completely taken over.

I think ideally we'd only need these two. This PR brings some flavor of the "low end" and another means called `ExportInterceptor`. The "high end" can come in the future when somebody gets around to it.

The `ExportInterceptor` is a somewhat narrow interception mechanism that doesn't really seem to have any use case beyond catching the error and selectively eating it away, and I'm feeling bit uneasy about a hook with narrow use case. @jglick had left a similar comment in this thread, too.

That said, I think the blue ocean project is an important enough "customer" that we can let this one in. I hate to make Vivek suffer more on this PR that he already had.

The few changes I propose are:
- the `skipIfFail` flag to become "as failure tolerant as possible" flag, which means it should also skip a failure in retrieving from a getter (this can be done by the default implementation of `ExportInterceptor` using this flag to change the behaviour, so that the custom `ExportInterceptor` can override this behaviour)
- `ExportInterceptor` returning null should produce the property with the `null` value, as opposed to skipping the property.
- `ExportInterceptor` defining a special constant that it can return to cause the property to skip.

vivek:
@kohsuke I have addressed your concerns, PTAL.

daniel-beck:
> narrow interception mechanism that doesn't really seem to have any use case beyond catching the error and selectively eating it away, and I'm feeling bit uneasy about a hook with narrow use case … I think the blue ocean project is an important enough "customer" that we can let this one in.

I must be misunderstanding this comment. Didn't we let a half dozen broken changes in, over the last year, for BO? Isn't it time BO driven changes to Stapler improve, and acknowledge the fact that most things that get added to Stapler or Jenkins will be around forever as a maintenance burden?

kohsuke:
The code change looks good to me for merge. I'm reaching out to Daniel offline to discuss his concern.

kohsuke:
I talked with Daniel offline and I think he's OK with my perspective that the long term nature of his concern should be separated from this particular PR at hand.

kohsuke:
:bee:

vivek:
@oleg-nenashev your compatibility concern had been addressed, do I have your bee on this:) 
cc: @daniel-beck @jglick

vivek:
Didn't hear from others, looks like all good. Have a bee from Kohsuke. Proceeding with merge.

core/src/main/java/org/kohsuke/stapler/export/ExportConfig.java
```diff
@@ -38,4 +47,40 @@ public ExportConfig withClassAttribute(ClassAttributeBehaviour cab) {
         this.classAttribute = cab;
         return this;
     }
+
+    /**
+     * Controls serialization of {@link @Exported} properties.
```
jglick: Syntax error. The `Jenkinsfile` is supposed to be validating that. I am trying to figure out why the build did not fail.

core/src/main/java/org/kohsuke/stapler/StaplerResponse.java
```diff
@@ -173,10 +175,30 @@
...
+     */
+    void serveExposedBean(StaplerRequest req, Object exposedBean, ExportConfig exportConfig) throws ServletException,IOException;
```
batmat: FWIW, this change is causing compile errors in some proprietary plugin. Wondering if we couldn't have leveraged a `default method` here to avoid it. 

Like for instance:

```java
serveExposedBean(req, exposedBean, config.getFlavor());
```

WDYT @jglick @vivek ?

jglick: cf. #128

PR #107, opened 2017-02-14
Handle CustomExportedBean serialization with Exported.merge=true

vivek:
CustomExportedBean serialization fails when a CustomExportedBean property is merged with @Exported(merge=true).

This was reported in https://issues.jenkins-ci.org/browse/JENKINS-41806.

```
org.kohsuke.stapler.export.NotExportableException: class org.kohsuke.stapler.export.ModelTest$Result doesn't have @ExportedBean so cannot write org.kohsuke.stapler.export.ModelTest$Build.buildResult

  at org.kohsuke.stapler.export.Model.<init>(Model.java:78)
  at org.kohsuke.stapler.export.ModelBuilder.get(ModelBuilder.java:51)
  at org.kohsuke.stapler.export.Property.writeTo(Property.java:141)
  at org.kohsuke.stapler.export.Model.writeNestedObjectTo(Model.java:226)
  at org.kohsuke.stapler.export.Model.writeTo(Model.java:197)
  at org.kohsuke.stapler.export.Model.writeTo(Model.java:217)
  at org.kohsuke.stapler.export.Model.writeTo(Model.java:181)
```

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/export/Model.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Model.java
@@ -23,6 +23,11 @@
 
 package org.kohsuke.stapler.export;
 
+import com.google.common.base.Predicate;
+import org.kohsuke.stapler.export.TreePruner.ByDepth;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Field;
@@ -36,12 +41,6 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nullable;
-
-import com.google.common.base.Predicate;
-import org.kohsuke.stapler.export.TreePruner.ByDepth;
 
 /**
  * Writes all the property of one {@link ExportedBean} to {@link DataWriter}.
--- a/core/src/main/java/org/kohsuke/stapler/export/Property.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Property.java
@@ -135,7 +135,8 @@ public abstract class Property implements Comparable<Property> {
         if (d==null && skipNull) { // don't write anything
             return;
         }
-        if (merge) {
+        //Merge doesn't apply to CustomExportedBean, so skip it
+        if (merge && !(d instanceof CustomExportedBean)) {
             // merged property will get all its properties written here
             if (d != null) {
                 Model model = owner.get(d.getClass(), parent.type, name);
--- a/core/src/test/java/org/kohsuke/stapler/export/ModelTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/export/ModelTest.java
@@ -23,14 +23,16 @@
  */
 package org.kohsuke.stapler.export;
 
-import static org.junit.Assert.assertEquals;
+import org.junit.Assert;
+import org.junit.Test;
 
+import java.io.IOException;
 import java.io.StringWriter;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 
-import org.junit.Test;
+import static org.junit.Assert.assertEquals;
 
 public class ModelTest {
     private ExportConfig config = new ExportConfig()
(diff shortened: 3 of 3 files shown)


i386:
@reviewbybees

vivek:
>Since we are returning a String, it has no properties—it is written as a JSON string, not a JSON object—so it would seem illegal to use @Exported(merge=true) on such a property, just as it would to do so on a property declared to return String. Probably the error message could be improved, but that is not what this PR is doing.

I guess thats arguable as it could also be a no-op, that is ignore merge in that case. 

> Anyway it is unclear how you think this would solve JENKINS-41806, since Build.getBuildResult does not use merge.

I could not reproduce JENKINS-41806, so that makes everything hard to pin point exactly what caused it. However the stack trace clearly points that this code path was traversed. See Property.java:140. you will get the same stack trace on Jenkins 2.45 if you try to serialize Result using merge=true. 

```
io.jenkins.blueocean.commons.stapler.export.NotExportableException: class hudson.model.Result doesn't have @ExportedBean so cannot write hudson.plugins.git.util.Build.buildResult
	at io.jenkins.blueocean.commons.stapler.export.Model.<init>(Model.java:80)
	at io.jenkins.blueocean.commons.stapler.export.ModelBuilder.get(ModelBuilder.java:53)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:190)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:169)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeTo(Property.java:140)
	at io.jenkins.blueocean.commons.stapler.export.Model.writeNestedObjectTo(Model.java:228)
```

> The @Exported property in question has been there since 2010, so the error seems to be specific to Blue Ocean. 

Right but merge property added sometime 2016 and blueocean wraps [action with merge=true](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueActionProxy.java#L17).

> From JIRA it is not clear whether you ever actually reproduced this.

I guess I need to reach out to submitter to help reproduce so that we are sure where exactly its coming from to be 100% sure.

jglick:
> blueocean wraps action with merge=true

That sounds like the culprit here. 

> the stack trace clearly points that this code path was traversed

Is it so clear? `writeTo` calls `writeValue`, the path shown at the end of the callstack, if and only if `!merge`. So a PR which makes this branch _more_ frequent would not fix the error—unless the critical call to `writeTo` is not the one you quoted, nor the one calling that one, but the calling _that_ one and delegating to `writeNestedObjectTo`:

```
io.jenkins.blueocean.commons.stapler.export.NotExportableException: class hudson.model.Result doesn't have @ExportedBean so cannot write hudson.plugins.git.util.Build.buildResult
	at io.jenkins.blueocean.commons.stapler.export.Model.<init>(Model.java:80)
	at io.jenkins.blueocean.commons.stapler.export.ModelBuilder.get(ModelBuilder.java:53)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:190)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:169)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeTo(Property.java:140)
	at io.jenkins.blueocean.commons.stapler.export.Model.writeNestedObjectTo(Model.java:228)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:280)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:169)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:244)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeValue(Property.java:169)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeTo(Property.java:140)
	at io.jenkins.blueocean.commons.stapler.export.Model.writeNestedObjectTo(Model.java:228)
	at io.jenkins.blueocean.commons.stapler.export.Property.writeTo(Property.java:136)
	at …
```

And the stack trace involves your forked copy of Stapler, not the one in Jenkins 2.45.

vivek:
>blueocean wraps action with merge=true
That sounds like the culprit here.

Its also important to note that this code does render action object only if that action object is annotated with @ExportedBean. So perhaps thats what causing it. 

>Is it so clear? writeTo calls writeValue, the path shown at the end of the callstack, if and only if !merge. So a PR which makes this branch more frequent would not fix the error—unless the critical call to writeTo is not the one you quoted, nor the one calling that one, but the calling that one and delegating to writeNestedObjectTo:

If I parse your statement correctly, then thats where fix was made. [See](https://github.com/vivek/stapler/blob/762221f1043a9d626e671d47f4f4886d3694c4c8/core/src/main/java/org/kohsuke/stapler/export/Property.java#L139)

> And the stack trace involves your forked copy of Stapler, not the one in Jenkins 2.45.

I copied it from JIRA, whatever user reported stack trace where it says Jenkins 2.45. Maybe user's env is mixed up. I am asking submitter for more info and get to validate it.

vivek:
Finally have been able to reproduce with latest instructions from submitter. This failure was happening due to imported CustomExportedBean in blueocean and this was failing instance check resulting in to this error. Its been fixed in blueocean. Closing this PR.

Issue #109, opened 2017-04-20
Quietly ignore interfaces in IndexHtmlDispatcher


varyvol:
After [PR #102](https://github.com/stapler/stapler/pull/102/files), the class hierarchy is inspected to look for index.html files [(make() method)](https://github.com/stapler/stapler/blob/0b91362e14498c02d1f8da1bf6f77fbacede6326/core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java#L39).

In case the used class is an interface, getSuperclass() returns null and thus a NPE happens. Although this probably is something conceptually wrong, it is something that worked previously and hence interfaces should be silently ignored for compatibility reasons.

oleg-nenashev:
@varyvol Could you please add a stacktrace to this issue?

varyvol:
@oleg-nenashev sure.

`org.apache.commons.jelly.JellyTagException: jar:file:/tmp/config-file-provider/WEB-INF/lib/config-file-provider.jar!/org/jenkinsci/plugins/configfiles/ConfigFilesManagement/configfiles.jelly:29:83: <st:include> java.lang.NullPointerException
  at org.apache.commons.jelly.impl.TagScript.handleException(TagScript.java:726)
  at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:281)
  at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
  at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
  at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
  at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
  at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
  at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
  at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
  at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
  at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
  at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
[...]

PR #110, opened 2017-04-20
Quietly ignore interfaces in IndexHtmlDispatcher, avoiding NPE.

varyvol:
[Issue #109](https://github.com/stapler/stapler/issues/109)

Interfaces do not have a superclass and thus a NPE exception was happening.

@reviewbybees @jglick @oleg-nenashev

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java
+++ b/core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java
@@ -33,7 +33,10 @@ class IndexHtmlDispatcher extends Dispatcher {
     }
 
     /**
-     * Returns a {@link IndexHtmlDispatcher} if and only if the said class has {@code index.html} as a side-file
+     * Returns a {@link IndexHtmlDispatcher} if and only if the said class has {@code index.html} as a side-file.
+     * @param context
+     * @param c Class from where to start the index.html inspection. Will go through class hierarchy (unless interface)
+     * @return Index dispatcher or {@code null} if it cannot be found
      */
     static Dispatcher make(ServletContext context, Class c) {
         for (; c != null && c != Object.class; c = c.getSuperclass()) {
--- a/core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java
+++ b/core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java
@@ -36,7 +36,7 @@ class IndexHtmlDispatcher extends Dispatcher {
      * Returns a {@link IndexHtmlDispatcher} if and only if the said class has {@code index.html} as a side-file
      */
     static Dispatcher make(ServletContext context, Class c) {
-        for (; c != Object.class; c = c.getSuperclass()) {
+        for (; c != null && c != Object.class; c = c.getSuperclass()) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java
```diff
@@ -36,7 +36,7 @@ public String toString() {
      * Returns a {@link IndexHtmlDispatcher} if and only if the said class has {@code index.html} as a side-file
      */
     static Dispatcher make(ServletContext context, Class c) {
```
oleg-nenashev: I would suggest documenting the null case in Javadoc/Annotations here and in Calling methods just to make sure nobody misuses the API.

varyvol:
@oleg-nenashev I might be wrong, but this does not ignore the interface per se but avoids the inspection of the class hierarchy, so my understanding is this should restore the previous behaviour.

core/src/main/java/org/kohsuke/stapler/IndexHtmlDispatcher.java
varyvol: Do you mean this method returning null? That should be a valid scenario in case no index.html is found.

oleg-nenashev: Nope, I just mean the documentation. 

E.g.

``` java
     /**
       * Returns a {@link IndexHtmlDispatcher} if and only if the said class has {@code index.html} as a side-file
       * @param c Class to be used. For {@code null} (e.g. interface) the returned value will be {@code null} as well
       * @return Index dispatcher or {@code null} if it cannot be found
       */
      @CheckForNull
      static Dispatcher make(ServletContext context, @CheckForNull Class c) {
```

oleg-nenashev:
@varyvol It is great if it fixes the issue :) Just misunderstood the PR then. Added the Javadoc enhancement proposal above. If you write a direct unit test for the case, it will be perfect

core/src/test/java/org/kohsuke/stapler/IndexHtmlDispatcherTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class IndexHtmlDispatcherTest {
+
+    @Test
+    @Issue("#109")
+    public void ignoreInterfaces() throws MalformedURLException {
+        ServletContext context = mock(ServletContext.class);
+        when(context.getResource(any(String.class))).thenReturn(null);
+
+        IndexHtmlDispatcher.make(context, TestInterface.class);
```
jglick: `assertNull` the result too.

varyvol:
@oleg-nenashev @jglick comments addressed.

oleg-nenashev:
@reviewbybees done

varyvol:
@oleg-nenashev @jglick do you know who might merge this?

oleg-nenashev:
@varyvol The merge can be done by several people, @vivek and @kohsuke are the maintainers AFAIK. I pinged them both on Friday regarding the release. We may want to also backport the fix to 2.46.3, but it will require a custom branch of Stapler (to be also decided by the maintainers). The release can be done by @kohsuke only.

jglick:
I am able to cut releases. Anyway this PR needs no immediate release, much less backport, since it is merely a robustness fix and the plugin bug was already corrected.

PR #112, opened 2017-04-24
JENKINS-40979  NPE where child can be null

i386:
Discovered that in certain conditions `child` can be null and cause an NPE https://github.com/jenkinsci/blueocean-plugin/pull/996


vivek:
LGTM 🐝

vivek:
cc:@reviewbybees

oleg-nenashev:
@reviewbybees done

jglick:
Also should have added nullability annotations to related parameters: `FilteringTreePruner.base`, `Model.writeNestedObjectTo#pruner`, `Property.writeTo#pruner`.

PR #113, opened 2017-05-01
Verifying that web methods may be defined as default interface methods

jglick:
Relevant since Jenkins core is now on `-source 8`.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <java.level>7</java.level>
+    <java.level>8</java.level>
   </properties>
 
   <build>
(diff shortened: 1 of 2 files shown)


jglick:
CI build is getting a SIGKILL for some reason.

jglick:
In fact [this build passed](https://jenkins.ci.cloudbees.com/job/core/job/stapler/178/) so it is just odd that there seem to be two jobs building the same repo. I will disable the other!

pom.xml
```diff
@@ -83,7 +83,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <java.level>7</java.level>
+    <java.level>8</java.level>
```
oleg-nenashev: Requires a decision by Stapler maintainers. LGTM, but may require extra backporting methods if somebody wants to just update LTS version to the Master branch. I don't

jglick: > Requires a decision by Stapler maintainers.

Well, I cut the last release…

> if somebody wants to just update LTS version to the Master branch

I think generally we do not do that. Rather, we would create a backport branch and cherry-pick selected fixes.

jglick:
@reviewbybees done

jglick:
but CC @kohsuke in case he has any opinion

jglick:
Will hold off on merge just in case.

PR #117, opened 2017-05-23
Make ClassDescriptor deterministic

jglick:
Has sometimes been observed that a web method with multiple overloads will randomly run one or the other (according to JVM implementation details), which is confusing and can lead to bugs, especially if all but one overload was intentionally deprecated. This patch simply picks one nondeprecated overload consistently so there are no surprises (as in https://github.com/junit-team/junit4/pull/293 for example).

@reviewbybees

### DIFF
--- a/core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
@@ -273,6 +273,29 @@ public class DispatcherTest extends JettyTestCase {
         assertEquals(1, requirePostOnBase.hit);
     }
 
+    public void testOverloads() throws Exception {
+        TextPage p = new WebClient().getPage(new URL(url, "overloaded/x"));
+        assertEquals("doX()", p.getContent().trim());
+    }
+    public final Object overloaded = new Overloaded();
+    public static class Overloaded {
+        public HttpResponse doX() {
+            return HttpResponses.plainText("doX()");
+        }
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
```diff
@@ -273,6 +273,30 @@ public void testRequirePostOnBase() throws Exception {
...
+        public HttpResponse doX(StaplerRequest req) {
+            return HttpResponses.plainText("doX(StaplerRequest)");
+        }
+        public HttpResponse doX(StaplerResponse rsp) {
+            return HttpResponses.plainText("doX(StaplerResponse)");
+        }
+        public HttpResponse doX(StaplerRequest req, StaplerResponse rsp) {
+            return HttpResponses.plainText("doX(StaplerRequest, StaplerResponse)");
+        }
+        @WebMethod(name = "x")
```
daniel-beck: So this means that a magically named method will have precedence over a `@WebMethod`? This seems counterintuitive.

jglick: Then deprecate your other overloads. Again, the exact ordering of (nondeprecated) methods is arbitrary, so long as it is predictable so any field errors can be reproduced by developers. If you have more than one, you are in the wrong.

daniel-beck:
+0.1

jglick:
@reviewbybees done

PR #118, opened 2017-05-31
Replace HttpResponses methods adding platform-specific newlines

### LINKED ISSUE #91: HttpResponse uses platform dependant line terminators.

jglick:
More compatible alternative to #92. Fixes #91.

@reviewbybees

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
+++ b/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
@@ -203,8 +203,9 @@ public class HttpResponses {
     }
 
     /**
-     * Serves the literal HTML.
+     * @deprecated Adds a platform-specific newline; prefer {@link #literalHtml}.
      */
+    @Deprecated
     public static HttpResponse html(final String literalHtml) {
         return new HttpResponse() {
             public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/kohsuke/stapler/HttpResponses.java
```diff
@@ -226,6 +243,21 @@ public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object nod
         };
     }
 
+    /**
+     * Serves a plain text response.
+     */
+    public static HttpResponse text(final String text) {
```
oleg-nenashev: `since` + entry in the changelog?

jglick: `@since` is basically useless for Jenkins developers. Will add a changelog entry when merging.

PR #119, opened 2017-06-19
Allow for either getOutputStream or getWriter to be used before sendRedirect is called

reversefold:


jglick:
Sounds right but can this be proven in a test? Is there a particular filed bug this is addressing?

reversefold:
This is related to a PR I wrote for the jenkins-artifactory-plugin: https://github.com/JFrogDev/jenkins-artifactory-plugin/pull/54/files Originally I was using getWriter() but I got a confusing error message about getWriter() already having been called. It took me a while to track down exactly what was going wrong and switch to getOutputStream(). This PR makes it so that either can be used before a redirect.

jglick:
OK. Putting this on hold until I have a moment to figure out if this is amenable to unit testing.

PR #121, opened 2017-07-07
[STAPLER-121] Initial proposed consolidated annotations for Stapler

stephenc:
See [this thread on the jenkins-dev mailing list](https://groups.google.com/forum/#!topic/jenkinsci-dev/UrVVT8wbHIE) for context.

@reviewbybees @jenkinsci/code-reviewers 

Here is the initial proposed annotations.

ToDo:

- [x] Wire up the `@StaplerPath`, `@StaplerRMI` and `@StaplerObject`, annotations so that they actually work (the `@StaplerGET` / `@StaplerContent` / etc annotations work out of the box because they build on top of the `@InterceptorAnnotation` meta-annotation. The `@StaplerFacet` and `@StaplerFragment` annotations are documentation only - retained with runtime retention as some consistency checks possible with these may need to be implemented via a unit test generator. 
- [ ] Implement some annotation processor based consistency checks
- [ ] Implement some automated unit test generator for the additional checks that cannot be performed by an annotation processor
  - [ ] Investigate whether the unit test generator could be implemented as an annotation processor
- [x] Tests for the new annotations
- [ ] Documentation

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/MetaClass.java
+++ b/core/src/main/java/org/kohsuke/stapler/MetaClass.java
@@ -23,9 +23,9 @@
 
 package org.kohsuke.stapler;
 
-import java.lang.annotation.Annotation;
 import net.sf.json.JSONArray;
 import org.apache.commons.io.IOUtils;
+import org.kohsuke.stapler.annotations.StaplerFacet;
 import org.kohsuke.stapler.annotations.StaplerObject;
 import org.kohsuke.stapler.annotations.StaplerPath;
 import org.kohsuke.stapler.bind.JavaScriptMethod;
@@ -107,14 +107,6 @@ public class MetaClass extends TearOffSupport {
         this.dispatchers.clear();
         KlassDescriptor<?> node = new KlassDescriptor(klass);
         // TODO abstract this test into Klass so that e.g. Ruby can have idiomatic
-        boolean staplerObject = false;
-        // TODO use clazz.getDeclaredAnnotation(StaplerObject.class) != null when Java 8 baseline
-        for (Annotation o: clazz.getDeclaredAnnotations()) {
-            if (o instanceof StaplerObject) {
-                staplerObject = true;
-                break;
-            }
-        }
 
         dispatchers.add(new DirectoryishDispatcher());
 
@@ -122,12 +114,7 @@ public class MetaClass extends TearOffSupport {
             dispatchers.add(new HttpDeletableDispatcher());
 
         FunctionList staplerPaths = node.methods.staplerPath();
-        FunctionList impliedPaths;
-        if (staplerObject) {
-            impliedPaths = FunctionList.emptyList();
-        } else {
-            impliedPaths = node.methods.nonStaplerPath();
-        }
+        FunctionList impliedPaths = node.methods.nonStaplerPath();
 
         for (Function f: staplerPaths.staplerMethod()) {
             for (String name: StaplerPath.Helper.getPaths(f)) {
@@ -209,7 +196,7 @@ public class MetaClass extends TearOffSupport {
                 for (String name: StaplerPath.Helper.getPaths(f)) {
                     dispatchers.add(new FieldSelectDispatcher(name, f));
                 }
-            } else if (!staplerObject) {
+            } else if (!StaplerObject.Helper.isObject(f.getDeclaringClass())) {
                 dispatchers.add(new FieldSelectDispatcher(f.getName(), f));
             }
         }
@@ -355,8 +342,10 @@ public class MetaClass extends TearOffSupport {
         // TODO: check if we can route to static resources
         // which directory shall we look up a resource from?
 
-        for (Facet f : webApp.facets) {
-            f.buildFallbackDispatchers(this, dispatchers);
+        if (!StaplerObject.Helper.isObject(clazz) || StaplerFacet.Helper.hasFallback(clazz)) {
+            for (Facet f : webApp.facets) {
+                f.buildFallbackDispatchers(this, dispatchers);
+            }
         }
 
         // check public selector methods <obj>.getDynamic(<token>,...)
--- a/core/src/main/java/org/kohsuke/stapler/MetaClass.java
+++ b/core/src/main/java/org/kohsuke/stapler/MetaClass.java
@@ -23,6 +23,7 @@
 
 package org.kohsuke.stapler;
 
+import java.lang.annotation.Annotation;
 import net.sf.json.JSONArray;
 import org.apache.commons.io.IOUtils;
 import org.kohsuke.stapler.annotations.StaplerObject;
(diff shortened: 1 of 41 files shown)


core/src/main/java/org/kohsuke/stapler/annotations/StaplerPaths.java
```diff
@@ -0,0 +1,42 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Annotation to allow a method / field to have multiple {@link StaplerPath} annotations.
+ */
+@Target({METHOD, FIELD})
+@Retention(RUNTIME)
+@Documented
+public @interface StaplerPaths {
```
jglick: If you update to Java 8 (cf. #113), you can mark this `@RepeatableAnnotation`.

stephenc: I think you mean that I can mark `@StaplerPath` with `@Repeatable(StaplerPaths.class)` (https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Repeatable.html) which you will see I already have (though I am relying on compiling with Java 8 targetting Java 7 for the build right now)

jglick: Interestingly `enforceBytecodeVersion` is not complaining. MENFORCER bug, or as designed? At any rate, after #113 this is a nonissue.

stephenc:
@jglick so functionally the annotations are complete.

I still need to:

- [ ] Write some tests for annotated fields in complex class hierarchy
- [ ] Write some tests for facets in complex class hierarchy with different `@StaplerFacet` annotations present
- [ ] Write some more unit tests for the helper classes

That would close out the testing story... after that there is implementing some annotation processors and fleshing out the documentation.

core/src/main/java/org/kohsuke/stapler/annotations/StaplerFacet.java
```diff
@@ -0,0 +1,192 @@
...
+        /**
+         * Collects the {@link StaplerFacet} annotations of a class. Walks the class hierarchy and all implemented
+         * interfaces to collect the facets.
+         *
+         * @param clazz the class.
+         * @return the annotations.
+         */
+        public static List<StaplerFacet> facetsOf(Class<?> clazz) {
+            Map<String, StaplerFacet> result = new LinkedHashMap<>();
+            Stack<Class<?>> interfaces = new Stack<>();
```
Wadeck: `Stack` has a layer of synchronization that does not seem to be required in such context. What not simply using a `Deque` (with `ArrayDeque` impl.) instead to avoid the sync overload ?

core/src/main/java/org/kohsuke/stapler/annotations/StaplerMethodInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+            } else if (a instanceof StaplerMethod) {
+                String value = ((StaplerMethod) a).value();
+                if (value.equals(method) || value.equals(StaplerMethod.ALL)) {
+                    return true;
+                }
+            } else {
+                Class<? extends Annotation> t = a.annotationType();
+                InterceptorAnnotation ia = t.getAnnotation(InterceptorAnnotation.class);
+                if (ia != null && ia.value() == StaplerMethodInterceptor.class) {
+                    if (t.getSimpleName().equals("Stapler" + method)) {
```
Wadeck: WDYT of using meta-annotation instead of string concat to check the method ? Something similar to the `Implicit`.

It's a bit weird to use the name of the class for such check.

core/src/main/java/org/kohsuke/stapler/annotations/StaplerPath.java
```diff
@@ -0,0 +1,372 @@
...
+                if (a instanceof StaplerPaths) {
+                    hasPathAnnotation = true;
+                    StaplerPath[] paths = ((StaplerPaths) a).value();
+                    if (paths.length == 0) {
+                        hasInferredName = true;
+                    } else {
+                        for (StaplerPath p: paths) {
+                            switch (p.value()) {
+                                case DYNAMIC:
+                                    continue;
```
Wadeck: it's weird to use a `continue` there that will jump back to the `for loop` 10 lines above but the next `break`s are relative to the `switch`. Not intuitive to read.

core/src/main/java/org/kohsuke/stapler/annotations/package-info.java
```diff
@@ -0,0 +1,117 @@
...
+ * {@link org.kohsuke.stapler.annotations}-based annotations and therefore more complete consistency checks can
+ * be enforced at compile time.
+ * <p>
+ * TODO: determine how to handle inheritance
+ * </li>
+ * </ul></li>
+ * </ul>
+ * @since TODO
+ */
+package org.kohsuke.stapler.annotations;
```
Wadeck: Nice use of the package-info to put some "package-scope" documentation 👍

core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
```diff
@@ -359,6 +989,4 @@ public void testPublicFieldDispatch() throws Exception {
     }
 
 
-
-
-}
\ No newline at end of file
+}
```
Wadeck: Very nice set of tests here. Perhaps just split into multiple class to ease the readability ?

core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
```diff
@@ -80,13 +109,248 @@ private void check(WebClient wc, HttpMethod m) throws java.io.IOException {
...
+    }
+
+    /**
+     * Tests the dispatching of StaplerPath based on method
+     */
+    public void testMethodMatch() throws Exception {
+        String[] methods = {"CONNECT", "DELETE", "GET", "HEAD", "PATCH", "POST", "PUT", "CUSTOM"};
+        for (String method1 : methods) {
+            check(method1, "all", 200);
+            // check(method1, "", 200); // FIXME uncomment once StaplerPath support added
```
stephenc: note to self... uncomment?

core/src/main/java/org/kohsuke/stapler/MetaClass.java
```diff
@@ -175,124 +192,94 @@ public String toString() {
...
+                dispatchers.add(new LongArgSelectDispatcher(name, f.contextualize(new TraversalMethodContext(name))));
+            }
         }
 
-        FunctionList getMethods = node.methods.prefix("get");
+        FunctionList getMethods = impliedPaths.prefix("get");
 
         // check public selector methods of the form NODE.getTOKEN()
         for (Function f : getMethods.signature()) {
-            if(f.getName().length()<=3)
```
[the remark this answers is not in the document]
stephenc: <= will only ever match in == but is a "safer" check in case somebody changes something else somewhere above IIUC

stephenc:
@daniel-beck is this irrelevant now?

jglick:
I do not think it is irrelevant, but it would require rework. CC @Wadeck

Wadeck:
This PR proposed additional annotations that still could be interesting but not in the current status. The source code was modified in depth for the security release and so this PR should be closed and start from the current version if we really want to have it included.

stephenc:
I'm going to close this so. If somebody wants to pick it up they can take on the rebase

jvz:
Some of this may be useful towards fixing [SECURITY-534](https://issues.jenkins-ci.org/browse/SECURITY-534) which I'm working on now.

PR #122, opened 2017-07-27
JENKINS-45099 special JSON characters are not consistently quoted

i386:
Special JSON characters such as `[` and `]` were not consistently quoted when writing to the output stream, resulting in invalid JSON.

This change uses `commons-lang` v3's `org.apache.commons.lang3.StringEscapeUtils#escapeJson` which ensures that all special JSON characters and control characters are encoded correctly. 

PTAL @reviewbybees @jglick @vivek

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -47,6 +47,11 @@
       <version>1.7.0</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.5</version>
+    </dependency>
+    <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
       <version>1.3.1</version>
--- a/core/src/test/java/org/kohsuke/stapler/export/JSONDataWriterTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/export/JSONDataWriterTest.java
@@ -5,6 +5,8 @@ import java.io.StringWriter;
 import java.util.Arrays;
 import java.util.List;
 
+import net.sf.json.JSONObject;
+import net.sf.json.util.JSONUtils;
 import org.junit.Test;
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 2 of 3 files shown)


michaelneale:
LGTM

stephenc:
Would be great if this was as two commits to show the failing test on CI first and then show how this change fixes... because I'm not seeing the encoding issue in the test case

I would also expect tests covering Unicode escaping that escapeJson adds to the mix

core/pom.xml
```diff
@@ -47,6 +47,11 @@
       <version>1.7.0</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.5</version>
```
oleg-nenashev: Jenkins widely uses 3.x: https://github.com/search?l=Maven+POM&p=2&q=org%3Ajenkinsci+%22commons-lang3%22&type=Code . The dependency should be fine

i386:
@stephenc I added the extra tests you asked for.

core/pom.xml
i386: could always shade it in to be safe.

core/src/test/java/org/kohsuke/stapler/export/JSONDataWriterTest.java
```diff
@@ -18,6 +18,31 @@
...
+        public String getValue() {
+            return "\"'/[]{}\\";
+        }
+    }
+
+    @ExportedBean
+    public static class JsonCharacters2 {
+        @Exported
+        public String getValue() {
+            return "\"foo\" isn't \"bar\". specials: \b\r\n\f\t\\/";
```
stephenc: I'm not seeing any tests covering the change this brings to escaping Unicode characters (specifically the method you are using will use the `\u0080` for any codepoint above 0x80... not strictly necessary for JSON as that should be in a UTF encoding always

michaelneale:
ping @i386

i386:
@michaelneale I don't think this is a quick fix and probably a suckhole for my time.

i386:
@stephenc thanks for the review feedback. Ill find someone else to look at this.

michaelneale:
@i386 right - does https://github.com/jenkinsci/blueocean-plugin/pull/1281 need to be undone on the blue side of things now or is that different? 

This seems to be a deep yak hole

PR #123, opened 2017-07-31
[JENKINS-41631] Lower Guava dep to 11, which is what Jenkins still uses

jglick:
Was noted in https://github.com/jenkinsci/plugin-pom/pull/67.

@reviewbybees

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>14.0</version>
+      <version>11.0.1</version>
     </dependency>
     <dependency>
       <groupId>org.kohsuke</groupId>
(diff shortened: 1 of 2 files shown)


jglick:
> Stapler is being used in many libs embedded into Jenkins libs and plugins

Such as?

> some of them may depend on Guava API available in newer versions

Clearly not if they were being used from Jenkins. If Stapler is being used from some non-Jenkins application, that application is free to depend on any newer version it likes.

> Same for non-Jenkins components like IDE plugins.

Huh? IDE plugins do not depend on Stapler. Why would they?

> Maybe we could "just" update Jenkins instead.

That would be far, far riskier. Which is why I am doing the safe fix now. If you want to take on updating Guava in Jenkins core and analyzing the impact of _that_, go ahead. (And it is not mutually exclusive with this patch anyway.)

jglick:
@oleg-nenashev ping

PR #127, opened 2017-08-16
[Performance] Export Model constructor throws exceptions on very hot paths

i386:
Removes additional millis per call on large exports by avoiding throwing `NotExportableException` and instead using annotation check to check exportability e.g. `type.getAnnotation(ExportedBean.class)`

See https://github.com/jenkinsci/blueocean-plugin/pull/1323

@reviewbybees #NightTimeCode

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
@@ -42,12 +42,22 @@ public class ModelBuilder {
     /*package*/ final Map<Class, Model> models = new ConcurrentHashMap<Class, Model>();
 
     public <T> Model<T> get(Class<T> type) throws NotExportableException {
+        if (type.getAnnotation(ExportedBean.class) == null) {
+            throw new NotExportableException(type);
+        }
         return get(type, null, null);
     }
 
-    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
+    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
+        if (type.getAnnotation(ExportedBean.class) == null) {
+            throw new NotExportableException(type);
+        }
+        return getOrNull(type, propertyOwner, property);
+    }
+
+    public <T> Model<T> getOrNull(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
         Model m = models.get(type);
-        if(m==null) {
+        if(m==null && type.getAnnotation(ExportedBean.class) != null) {
             m = new Model<T>(this, type, propertyOwner, property);
         }
         return m;
--- a/core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
@@ -47,6 +47,10 @@ public class ModelBuilder {
         return get(type, null, null);
     }
 
+    /**
+     * @throws NotExportableException if type is not exportable
+     * @return model
+     */
     @Nonnull
     public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
         Model<T> model = getOrNull(type, propertyOwner, property);
@@ -56,6 +60,12 @@ public class ModelBuilder {
         return model;
     }
 
+    /**
+     * Instead of throwing {@link NotExportableException} this method will return null
+     * This should be used on hot paths where throwing the exception and catching it would incur a performance hit
+     * @return model
+     * @since 1.253
+     */
     @CheckForNull
     public <T> Model<T> getOrNull(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
         Model<T> m = models.get(type);
--- a/core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
@@ -26,6 +26,7 @@ package org.kohsuke.stapler.export;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 /**
@@ -36,27 +37,28 @@ import javax.annotation.Nullable;
 public class ModelBuilder {
     /**
      * Instantiated {@link Model}s.
-     * Registration happens in {@link Model#Model(ModelBuilder,Class)} so that cyclic references
+     * Registration happens in {@link Model#Model(ModelBuilder, Class, Class, String)} so that cyclic references
      * are handled correctly.
      */
     /*package*/ final Map<Class, Model> models = new ConcurrentHashMap<Class, Model>();
 
+    @Nonnull
     public <T> Model<T> get(Class<T> type) throws NotExportableException {
-        if (type.getAnnotation(ExportedBean.class) == null) {
-            throw new NotExportableException(type);
-        }
         return get(type, null, null);
     }
 
-    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
-        if (type.getAnnotation(ExportedBean.class) == null) {
+    @Nonnull
+    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
+        Model<T> model = getOrNull(type, propertyOwner, property);
+        if (model == null) {
             throw new NotExportableException(type);
         }
-        return getOrNull(type, propertyOwner, property);
+        return model;
     }
 
-    public <T> Model<T> getOrNull(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
-        Model m = models.get(type);
+    @CheckForNull
+    public <T> Model<T> getOrNull(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
+        Model<T> m = models.get(type);
         if(m==null && type.getAnnotation(ExportedBean.class) != null) {
             m = new Model<T>(this, type, propertyOwner, property);
         }
--- a/core/src/main/java/org/kohsuke/stapler/export/Property.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Property.java
@@ -198,6 +198,12 @@ public abstract class Property implements Comparable<Property> {
 
         Class c = value.getClass();
 
+        try {
+            writer.type(expected, value.getClass());
+        } catch (AbstractMethodError _) {
+            // legacy impl that doesn't understand it
+        }
+
         if (c.getAnnotation(ExportedBean.class) == null) {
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -199,118 +198,127 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
             }
-            if(value instanceof Iterable) {
-                writer.startArray();
-                Type expectedItemType = Types.getTypeArgument(expected, 0, null);
-                for (Object item : pruner.getRange().apply((Iterable) value)) {
-                    writeBuffered(expectedItemType, item, pruner, writer);
+        }
+
+        try {
+            writer.type(expected, value.getClass());
```
vivek: This looks like 🐛 `writer.type()` should be called before writing the object on line 209.

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -199,118 +198,127 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
-                    for (Object item : r.apply((Object[]) value)) {
-                        writeBuffered(act, item, pruner, writer);
-                    }
-                } else {
-                    // more generic case
-                    int len = Math.min(r.max, Array.getLength(value));
-                    for (int i=r.min; i<len; i++) {
-                        writeBuffered(act, Array.get(value, i), pruner, writer);
-                    }
+            writer.type(expected, value.getClass());
```
vivek: It should move down inside else condition below, like 211. handleNotExportable has its own logic to write type.

i386: right you are. Moved.

i386:
Stapler passed locally.

i386:
@stephenc @jglick @daniel-beck a review here would be fantastic please :)

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -42,12 +42,22 @@
     /*package*/ final Map<Class, Model> models = new ConcurrentHashMap<Class, Model>();
 
     public <T> Model<T> get(Class<T> type) throws NotExportableException {
+        if (type.getAnnotation(ExportedBean.class) == null) {
+            throw new NotExportableException(type);
```
jglick: Is this not redundant? AFAICT the method it delegates to does the same thing.

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -42,12 +42,22 @@
...
+        }
         return get(type, null, null);
     }
 
-    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
```
jglick: Why removing the `throws` clause?

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -42,12 +42,22 @@
...
+    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
+        if (type.getAnnotation(ExportedBean.class) == null) {
```
jglick: I think you rather meant to write

```java
Model<T> model = getOrNull(type, propertyOwner, property);
if (model == null) {
    throw new NotExportableException(type);
}
return model;
```

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -42,12 +42,22 @@
...
 
     public <T> Model<T> get(Class<T> type) throws NotExportableException {
+        if (type.getAnnotation(ExportedBean.class) == null) {
+            throw new NotExportableException(type);
+        }
         return get(type, null, null);
     }
 
-    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
+    public <T> Model<T> get(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
```
jglick: `@Nonnull` return

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -42,12 +42,22 @@
...
+        if (type.getAnnotation(ExportedBean.class) == null) {
+            throw new NotExportableException(type);
+        }
+        return getOrNull(type, propertyOwner, property);
+    }
+
+    public <T> Model<T> getOrNull(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) throws NotExportableException {
```
jglick: `@CheckForNull` return. And why is this declared to throw `NotExportableException`?

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -131,23 +131,22 @@ public void writeTo(Object object, TreePruner pruner, DataWriter writer) throws
         TreePruner child = pruner.accept(object, this);
         if (child==null)        return;
 
-        Object d = writer.getExportConfig().getExportInterceptor().getValue(this,object, writer.getExportConfig());
+        Object d = writer.getExportConfig().getExportInterceptor().getValue(this,object,writer.getExportConfig());
```
jglick: gratuitous, revert

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -199,118 +198,126 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
+                Model model = owner.getOrNull(c, parent.type, name);
+                if (model == null) {
+                    throw new NotExportableException(c);
                 }
-                writer.endArray();
+                writer.startObject();
+                model.writeNestedObjectTo(value, pruner, writer);
+                writer.endObject();
+            } catch (NotExportableException ex) {
+                handleNotExportable(expected, value, pruner, writer, skipIfFail, c);
```
jglick: [reduced diff](https://github.com/stapler/stapler/pull/127/files?w=1) is easier to read here (generally dislike patches that change indentation of large existing code blocks but sometimes it is hard to avoid)

i386:
@jglick thanks for the review. There was a fair bit of dumb code here on my part, so I did my best to fix it.

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -199,118 +197,122 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
+            writer.endArray();
+            return;
+        }
+        if(value instanceof Map) {
+            if (verboseMap!=null) {// verbose form
+                writer.startArray();
+                for (Map.Entry e : ((Map<?,?>) value).entrySet()) {
+                    BufferedDataWriter buffer = new BufferedDataWriter(writer.getExportConfig());
+                    try {
+                        writeStartObjectNullType(buffer);
```
jglick: Hard to follow this long diff, but IIUC you are basically just extracting `handleNotExportable` and then calling it twice, right? Pretty much unreviewable so I hope the test coverage here is good.

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -199,118 +197,122 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
         Class c = value.getClass();
 
-        Model model;
-        try {
-            model = owner.get(c, parent.type, name);
-        } catch (NotExportableException ex) {
-            if(STRING_TYPES.contains(c)) {
-                writer.value(value.toString());
-                return;
+        if (c.getAnnotation(ExportedBean.class) == null) {
```
jglick: Am I missing something or would this not be caught by the null return value from `getOrNull` a few lines later anyway?

core/src/main/java/org/kohsuke/stapler/export/Property.java
jglick: Actually the [whitespace-reduced diff](https://github.com/stapler/stapler/pull/127/files?w=1) is more approachable; there must have been a change in indentation for a long block of code.

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -199,118 +197,122 @@ private void writeValue(Type expected, Object value, TreePruner pruner, DataWrit
...
-                } else {
-                    // more generic case
-                    int len = Math.min(r.max, Array.getLength(value));
-                    for (int i=r.min; i<len; i++) {
-                        writeBuffered(act, Array.get(value, i), pruner, writer);
-                    }
+        }
+    }
+
+    private void handleNotExportable(Type expected, Object value, TreePruner pruner, DataWriter writer, boolean skipIfFail, Class c) throws IOException {
```
jglick: The method name is potentially misleading. Really this method is handling anything without `@ExportedBean`, but that includes all sorts of objects which are “exportable”. It is only at the very end, starting where we check `skipIfFail`, that this method is in fact handling objects which are not exportable.

core/src/main/java/org/kohsuke/stapler/export/Property.java
```diff
@@ -139,16 +140,12 @@ public void writeTo(Object object, TreePruner pruner, DataWriter writer) throws
...
             if (d != null) {
-                Model model;
-                try {
-                    model = owner.get(d.getClass(), parent.type, name);
-                } catch (NotExportableException e) {
-                    if(writer.getExportConfig().isSkipIfFail()){
-                        return;
-                    }
-                    throw e;
+                Model model = owner.getOrNull(d.getClass(), parent.type, name);
```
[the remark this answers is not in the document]
i386: Done

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -36,18 +37,29 @@
...
-        if(m==null) {
+        Model<T> model = getOrNull(type, propertyOwner, property);
+        if (model == null) {
+            throw new NotExportableException(type);
+        }
+        return model;
+    }
+
+    @CheckForNull
+    public <T> Model<T> getOrNull(Class<T> type, @CheckForNull Class<?> propertyOwner, @Nullable String property) {
```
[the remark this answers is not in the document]
i386: Done.

core/src/main/java/org/kohsuke/stapler/export/Property.java
i386: You are absolutely right. Fixed.

core/src/main/java/org/kohsuke/stapler/export/Property.java
i386: Renamed and added a description.

core/src/main/java/org/kohsuke/stapler/export/Property.java
i386: actually, Im going to revert the method extraction. Still too long for my taste but should preserve the diff.

core/src/main/java/org/kohsuke/stapler/export/ModelBuilder.java
```diff
@@ -56,6 +60,12 @@
         return model;
     }
 
+    /**
+     * Instead of throwing {@link NotExportableException} this method will return null
```
jglick: BTW tip: terminate summary sentences with `.`.

PR #128, opened 2017-08-18
Use default method for better backward compatibility

batmat:
https://github.com/stapler/stapler/pull/106 added a new method in StaplerResponse, causing backward compatibility issue in a closed source plugin.

As Stapler bumped to Java 8, this change leverages default method to make it more compatible.

@reviewbybees esp. @jglick @vivek @jtnord

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/StaplerResponse.java
+++ b/core/src/main/java/org/kohsuke/stapler/StaplerResponse.java
@@ -195,8 +195,11 @@ public interface StaplerResponse extends HttpServletResponse {
      * in detail; see {@link NamedPathPruner#NamedPathPruner(String)} for details.
      *
      * <p> {@link ExportConfig} is passed by the caller to control serialization behavior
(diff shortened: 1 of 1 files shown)


batmat:
@reviewbybees done.

batmat:
@oleg-nenashev nope, but can file it if need be. If you think this is worth it. But no backport is needed anyway since 2.73 got chosen and this was introduced in 2.74.

jglick:
What about `DataWriter.getExportConfig`? cf. https://github.com/stapler/stapler/pull/106#discussion_r102306777

jglick:
I do not think this is urgent. No one should be implementing `StaplerResponse` directly, since you can instead subclass `StaplerResponseWrapper`.

jglick:
Also cf. `catch AbstractMethodError` in code which #127 refactors.

jglick:
Filed #129 for follow-ups.

Issue #130, opened 2017-09-20
remove gzip compression from stapler


jtnord:
Stapler adds some code to do gzip compression on the fly.

However there are various routes as pointed out [in the code](https://github.com/stapler/stapler/blob/d80e59aad717ea8bfde0d2e37d6cdef7eda5e75a/core/src/main/java/org/kohsuke/stapler/compression/CompressionFilter.java#L18-L33).

All modern containers (tomcat, weblogic, jetty) have support built in for doing compression which alleviates these tricks.

Therefore in the modern age, just remove any support in stapler for compression handling and make users user their servlet container appropriately if they desire the configuration.

jglick:
If you can patch Winstone to apply compression by default (or perhaps at an option—unclear you want it for clients on the same LAN) then sure, would be great to simplify the code here.

jtnord:
quote from the [jetty docs](http://www.eclipse.org/jetty/documentation/9.3.x/gzip-filter.html)..

> The Jetty GzipHandler is a compression handler that you can apply to any dynamic resource (servlet). It fixes many of the bugs in commonly available compression filters: it works with asynchronous servlets; it handles all ways to set content length. It has been tested with Jetty continuations and suspending requests.

my plan was to remote the support in stapler and then it is up to the end user using stapler to enable it (either via a load balancer, or configuration e.g. tomcat [compression option in connector](https://tomcat.apache.org/tomcat-8.5-doc/config/http.html#Standard_Implementation)  and to add the jetty servlet dynamically in winstone if a flag to disable compression it is not provided

PR #131, opened 2017-09-20
[stapler#114] Update to Servlet API 3.1.0

### LINKED ISSUE #114: stapler implements Servlet API 2.5 which is pretty old
Stapler implements only the servlet 2.5 API - yet many applications now use 3.1 (either directly or indirectly_

If you attempt to directly use parts of the 3.1 API on a `StaplerRequest` or `StaplerResponse` then weird/bad things could happen depending on what you get underneath..
I have not observed any issues yet - but if you get a `org.kohsuke.stapler.StaplerResponseWrapper` rather than a `org.kohsuke.stapler.ResponseImpl` then things are liable to blow up. the same probably applies to `StaplerRequest` sub classes.

Apparantly (although no details are confirmed), this is causing some issues seen in the wild on the [Jenkins project](https://issues.jenkins-ci.org/browse/JENKINS-37062?focusedCommentId=298534&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-298534)

jtnord:
Update stapler to use Servlet API 3.1.0 (and other servlet APIs coming from Java EE 7)

Fixes #114 

- [x] manual testing using the [example maven project](https://github.com/stapler/stapler/pull/131/files#diff-b22338ed154eaacfd72f6b6c12442625)
- [x] corresponding PR to Jenkins incorporating the change in a real project (jenkinsci/jenkins#3033)

Code could be cleaned up further with the inclusion of #130 

### Proposed Change Log Entry

* updated API to compiled against Java Servlet API 3.1.0, JSP API 2.3.0 and JSTL API 1.2.1
* This required a API signature change of `org.kohsuke.stapler.compression.FilterServletOutputStream`.
  * It is recommended that users of this class do not upgrade but replace the implementation as this call may be removed in a future version.

@reviewbybees

### DIFF
--- a/core/maven-example/pom.xml
+++ b/core/maven-example/pom.xml
@@ -13,24 +13,25 @@
     <dependency>
       <groupId>org.kohsuke.stapler</groupId>
       <artifactId>stapler-jsp</artifactId>
-      <version>1.144</version>
+      <!-- TODO update after releasing -->
+      <version>1.253-SNAPSHOT</version>
     </dependency>
     <dependency>
-      <groupId>javax.servlet</groupId>
-      <artifactId>jstl</artifactId>
-      <version>1.2</version>
+      <groupId>javax.servlet.jsp.jstl</groupId>
+      <artifactId>javax.servlet.jsp.jstl-api</artifactId>
+      <version>1.2.1</version>
     </dependency>
 
     <dependency>
       <groupId>javax.servlet</groupId>
-      <artifactId>servlet-api</artifactId>
-      <version>2.3</version>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>3.1.0</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>javax.servlet</groupId>
-      <artifactId>jsp-api</artifactId>
-      <version>2.0</version>
+      <groupId>javax.servlet.jsp</groupId>
+      <artifactId>javax.servlet.jsp-api</artifactId>
+      <version>2.3.0</version>
       <scope>provided</scope>
     </dependency>
   </dependencies>
@@ -45,8 +46,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
 
         <configuration>
-          <source>1.5</source>
-          <target>1.5</target>
+          <source>1.7</source>
+          <target>1.7</target>
           <debug>true</debug>
           <showDeprecation>true</showDeprecation>
           <showWarnings>true</showWarnings>
@@ -61,28 +62,20 @@
       </plugin>
 
       <plugin>
-        <groupId>org.mortbay.jetty</groupId>
-        <artifactId>maven-jetty-plugin</artifactId>
-
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-maven-plugin</artifactId>
+        <version>9.4.7.v20170914</version>
         <configuration>
           <scanIntervalSeconds>2</scanIntervalSeconds>
 
-          <connectors>
-            <connector implementation="org.mortbay.jetty.nio.SelectChannelConnector">
-              <port>9090</port>
-              <maxIdleTime>60000</maxIdleTime>
-            </connector>
-          </connectors>
+          <httpConnector>
+            <host>localhost</host>
+            <port>9090</port>
+            <idleTimeout>60000</idleTimeout>
+          </httpConnector>
         </configuration>
       </plugin>
     </plugins>
   </build>
-
-  <repositories>
-    <repository>
-      <id>java.net2</id>
-      <url>http://download.java.net/maven/2/</url>
-    </repository>
-  </repositories>
 
 </project>
--- a/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
@@ -305,7 +305,7 @@ public class ResponseImpl extends HttpServletResponseWrapper implements StaplerR
         // CompressionFilter not available, so do it on our own.
         // see CompressionFilter for why this is not desirable
         setHeader("Content-Encoding","gzip");
-        return recordOutput(new FilterServletOutputStream(new GZIPOutputStream(super.getOutputStream())));
+        return recordOutput(new FilterServletOutputStream(new GZIPOutputStream(super.getOutputStream()), super.getOutputStream()));
     }
 
     public Writer getCompressedWriter(HttpServletRequest req) throws IOException {
--- a/core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
+++ b/core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
@@ -15,6 +15,7 @@ import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.io.Writer;
 import java.net.URL;
+import java.util.Collection;
 import java.util.Locale;
 
 /**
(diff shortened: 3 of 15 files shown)


jtnord:
fixes #114

core/maven-example/pom.xml
```diff
@@ -45,8 +46,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
 
         <configuration>
-          <source>1.5</source>
-          <target>1.5</target>
+          <source>1.7</source>
```
jtnord: minimum target of java ee 7

core/maven-example/pom.xml
```diff
@@ -13,24 +13,25 @@
     <dependency>
       <groupId>org.kohsuke.stapler</groupId>
       <artifactId>stapler-jsp</artifactId>
-      <version>1.144</version>
+      <!-- TODO update after releasing -->
+      <version>1.253-SNAPSHOT</version>
```
jtnord: not part of the build so would need to be fixed later.

core/maven-example/pom.xml
```diff
@@ -61,28 +62,20 @@
...
+            <host>localhost</host>
+            <port>9090</port>
+            <idleTimeout>60000</idleTimeout>
+          </httpConnector>
         </configuration>
       </plugin>
     </plugins>
   </build>
 
-  <repositories>
```
jtnord: stapler is now on maven central

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
```diff
@@ -305,7 +305,7 @@ public OutputStream getCompressedOutputStream(HttpServletRequest req) throws IOE
         // CompressionFilter not available, so do it on our own.
         // see CompressionFilter for why this is not desirable
         setHeader("Content-Encoding","gzip");
-        return recordOutput(new FilterServletOutputStream(new GZIPOutputStream(super.getOutputStream())));
+        return recordOutput(new FilterServletOutputStream(new GZIPOutputStream(super.getOutputStream()), super.getOutputStream()));
```
jtnord: this is pretty ugly - there is no need for stapler to do gzip at all - see #130

jtnord: this is ugly.

core/src/main/java/org/kohsuke/stapler/StaplerResponseWrapper.java
```diff
@@ -340,4 +341,46 @@ public void setContentType(String type) {
     public void setLocale(Locale loc) {
         getWrapped().setLocale(loc);
     }
+
+    /** {@inheritDoc} */
```
jtnord: new APIs in `ServletResponse`

core/src/main/java/org/kohsuke/stapler/compression/CompressionServletResponse.java
```diff
@@ -58,7 +58,7 @@ public ServletOutputStream getOutputStream() throws IOException {
     public void activate() throws IOException {
         if (stream==null) {
             super.setHeader("Content-Encoding", "gzip");
-            stream = new FilterServletOutputStream(new GZIPOutputStream(super.getOutputStream()));
+            stream = new FilterServletOutputStream(new GZIPOutputStream(super.getOutputStream()), stream);
```
jtnord: this is pretty ugly - there is no need for stapler to do gzip at all - see #130

core/src/main/java/org/kohsuke/stapler/compression/FilterServletOutputStream.java
```diff
@@ -11,9 +14,16 @@
...
-
-    public FilterServletOutputStream(OutputStream out) {
+    private final ServletOutputStream realSream;
+    
+    /**
+     * Constructs a new {@link FilterOutputStream}.
+     * @param out the stream that sits above the realStream, performing some filtering.  This must be eventually delegating eventual writes to {@code realStream}.
+     * @param realStream the actual underlying ServletOutputStream from the container.  Used to check the {@link #isReady()} state and to add {@link WriteListener}s. 
+     */
+    public FilterServletOutputStream(OutputStream out, ServletOutputStream realStream) {
```
jtnord: this is pretty ugly - there is no need for stapler to do gzip at all - see #130

jtnord: The usage of this in Jenkins core shows that this is not appropriate there.
I found 2 external uses of this 
1 in jenkins core
1 in a propriatary plugin

in both cases it is a mock stream, that does not respond to the client at all and as such has a different use case to this.

core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
jtnord: currently on 2 uses outside of stapler that I found - I have PRs for both, both of which means this can be completely removed in #130

core/maven-example/pom.xml
oleg-nenashev: works for us

core/maven-example/pom.xml
```diff
@@ -61,28 +62,20 @@
       </plugin>
 
       <plugin>
-        <groupId>org.mortbay.jetty</groupId>
-        <artifactId>maven-jetty-plugin</artifactId>
-
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-maven-plugin</artifactId>
+        <version>9.4.7.v20170914</version>
```
oleg-nenashev: CC @olamy just in case he is aware about any issues with that

olamy: nope

jtnord:
> Have you checked compatibility of other projects within the Stapler organization? Maybe they require update before merging

@oleg-nenashev No, updated this repo and the test project only.  I did search for the class that had an API signature change and found only 2 occurences  of it - neither of which where in the stapler org.  If you are aware of any other tests I should do can you outline them?  (I ran the example project `mvn jetty:run` and verified it started up and returned information for various pages)

oleg-nenashev:
@jtnord It should be fine. It would be great if you reference the changed methods in the PR description, so that we could mention it in the changelogs and upgrade guidelines if anything goes wrong.

jtnord:
@oleg-nenashev updated the PR description to add a proposed change-log

core/maven-example/pom.xml
batmat: I think I remember a discussion somewhere: is stapler stil in-use outside Jenkins? If not, and even if so, why not go to Java 8? It's probably time anyway, as now JDK9 just got GA a few days ago, even JDK8 is legacy now ;-).

jtnord: It is in a few places (2 or 3 iirc)  7 is required for the servlet spec which is the min requirement here

jglick: Stapler is on Java 8 already.

core/maven-example/pom.xml
jglick: FWIW we are using `9.4.5.v20170502` in various other places in the Jenkins project.

groovy/pom.xml
```diff
@@ -23,8 +23,8 @@
     </dependency>
     <dependency>
       <groupId>javax.servlet</groupId>
-      <artifactId>servlet-api</artifactId>
-      <version>2.3</version>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>3.1.0</version>
```
jglick: Please use `<dependencyManagement>` to avoid repetition.

PR #133, opened 2017-10-10
Resolve Upper Bound issues, update Groovy

oleg-nenashev:
- [x] - Update parent POM
- [x] - Add upper bound enforcement
- [x] - cleanup issues (mostly commons-*)
- [x] - Update Groovy to 2.4.11, so it is the same as the core baseline

All versions are compatible with the Jenkins core.

@reviewbybees @jglick

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -44,12 +44,22 @@
     <dependency>
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
-      <version>1.7.0</version>
+      <version>1.8.0</version>
     </dependency>
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>1.3.1</version>
+      <version>2.4</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
+      <version>1.2</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.9</version>
     </dependency>
     <dependency>
       <groupId>org.jvnet.localizer</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -149,7 +149,7 @@
           <dependency>
             <groupId>org.codehaus.mojo</groupId>
             <artifactId>extra-enforcer-rules</artifactId>
-            <version>1.0-beta-2</version>
+            <version>1.0-beta-4</version>
           </dependency>
         </dependencies>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.kohsuke</groupId>
     <artifactId>pom</artifactId>
-    <version>14</version>
+    <version>19</version>
   </parent>
   <groupId>org.kohsuke.stapler</groupId>
   <artifactId>stapler-parent</artifactId>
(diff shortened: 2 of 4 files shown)


core/pom.xml
```diff
@@ -44,12 +44,22 @@
     <dependency>
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
-      <version>1.7.0</version>
+      <version>1.8.0</version>
```
jglick: Are these aligned with `jenkins-core`?

pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.kohsuke</groupId>
     <artifactId>pom</artifactId>
-    <version>14</version>
+    <version>19</version>
```
oleg-nenashev: Ideally we should start migrating it to https://github.com/jenkinsci/pom.
IIUC there is no other things excepting Plugin site release and code signing there. WDYT @jglick ?

jglick:
Oops:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.3:site (default-site) on project stapler-parent: failed to get report for org.apache.maven.plugins:maven-javadoc-plugin: Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (default) on project stapler: Execution default of goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce failed. NullPointerException -> [Help 1]
```

core/pom.xml
oleg-nenashev: Jenkins core is at 1.8.3. Can bump it even more

oleg-nenashev:
Works on my machine (c). No sure what happens there

pom.xml
jglick: Not sure. Technically this is a non-Jenkins component.

oleg-nenashev: It's something we could also fix. Stapler is a critical part of Jenkins && it's not extremely  popular elsewhere, so moving it && integrating common Jenkins QA/review flow would not hurt. It's something we were discussing multiple times, but maybe it belongs to the ML now.

jglick:
Reproducible locally:

```
java.lang.NullPointerException
	at org.apache.maven.plugins.enforcer.EnforceBytecodeVersion.isBadArtifact(EnforceBytecodeVersion.java:195)
	at org.apache.maven.plugins.enforcer.EnforceBytecodeVersion.checkDependencies(EnforceBytecodeVersion.java:182)
	at org.apache.maven.plugins.enforcer.EnforceBytecodeVersion.handleArtifacts(EnforceBytecodeVersion.java:109)
	at org.apache.maven.plugins.enforcer.AbstractResolveDependencies.execute(AbstractResolveDependencies.java:77)
	at org.apache.maven.plugins.enforcer.EnforceMojo.execute(EnforceMojo.java:202)
```

You just need to update `extra-enforcer-rules` to `1.0-beta-4` or later to pick up the fix of https://github.com/mojohaus/extra-enforcer-rules/issues/13.

oleg-nenashev:
@jglick ptal

pom.xml
jglick: Basically that would mean not releasing it to Central any more. Fine with me, but needs some discussion.

jglick:
It seems `NumberConverter` regressed sometime between 1.7.0 and 1.8.0. 1.8.3 is also affected. But 1.9.3 is even worse—it does not convert `String[]`s properly either. Any I think I have identified a bug in how Stapler was calling BeanUtils that is responsible. Will file a follow-up PR.

PR #135, opened 2017-12-10
[JENKINS-34254] Allow customizing RequirePOST error pages

daniel-beck:
I'm pretty sure `requestURL` is mostly redundant, but cannot really hurt either to make it easily accessible.

Downstream use: https://github.com/jenkinsci/jenkins/pull/3187

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
+++ b/core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
@@ -49,7 +49,7 @@ public @interface RequirePOST {
         public Object invoke(StaplerRequest request, StaplerResponse response, Object instance, Object[] arguments)
                 throws IllegalAccessException, InvocationTargetException, ServletException {
             if (!request.getMethod().equals("POST")) {
-                for (ErrorCustomizer handler : ServiceLoader.load(ErrorCustomizer.class)) {
+                for (ErrorCustomizer handler : ServiceLoader.load(ErrorCustomizer.class, request.getWebApp().getClassLoader())) {
                     ForwardToView forwardToView = handler.getForwardView();
                     if (forwardToView != null) {
                         throw new InvocationTargetException(forwardToView.with("requestURL", request.getRequestURLWithQueryString().toString()));
--- a/core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
+++ b/core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
@@ -14,6 +14,8 @@ import java.util.ServiceLoader;
 
 import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.*;
+
+import javax.annotation.CheckForNull;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
```diff
@@ -31,11 +33,35 @@
...
+
+    /**
+     * Allows customizing the error page shown when an annotated method is called with the wrong HTTP method.
+     */
+    interface ErrorHandler {
+        ForwardToView getForwardView();
+    }
+
     public static class Processor extends Interceptor {
+        private static ErrorHandler handler;
```
oleg-nenashev: Maybe YAGNI, but it seems natural to have an option to register multiple handlers

core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
```diff
@@ -31,11 +33,35 @@
 @Target({METHOD,FIELD})
 @InterceptorAnnotation(RequirePOST.Processor.class)
 public @interface RequirePOST {
+
+    /**
+     * Allows customizing the error page shown when an annotated method is called with the wrong HTTP method.
+     */
+    interface ErrorHandler {
+        ForwardToView getForwardView();
```
oleg-nenashev: According to the code, it makes sense to document the `null` return here. `@CheckForNull`... you know

core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
```diff
@@ -31,11 +33,35 @@
...
+    }
+
     public static class Processor extends Interceptor {
+        private static ErrorHandler handler;
+
+        /**
+         * Register the custom error handler to be used when an annotated method is called with the wrong HTTP method.
+         * @param errorHandler the error handler providing the view to show instead
+         */
+        public static void registerErrorHandler(ErrorHandler errorHandler) {
```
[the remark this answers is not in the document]
daniel-beck: Could you provide an example? I assume that would use `ServiceLoader`, but I can't find references to that with the existing `@MetaInfServices` annotated classes.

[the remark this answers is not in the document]
jglick: Use `ServiceLoader`. I will file a separate PR to deal with the fact that you could not find it. :-)

daniel-beck:
@jglick Addressed your review comment and adapted downstream.

core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java
jglick: Agreed, and Javadoc here is the right place to discuss the first non-null value.

daniel-beck:
I think I now addressed everything.

daniel-beck:
Given the semantics change using `ServiceLoader` (multiple implementations) it could make sense to pass at least `StaplerRequest` to the `getForwardView()` method, but it can be obtained from `Stapler` anyway, and I don't expect that to be a common use case.

daniel-beck:
@jglick Would be great if you could merge and release this if you're happy with it, so it can be included in a weekly soon.

PR #136, opened 2018-01-18
Log when a property is left out of serialised result due to exception

sophistifunk:
Silently dropping properties has been causing issues in front-end code; this will add logging so we'll have something to look for when this happens in the future.

See: https://issues.jenkins-ci.org/browse/JENKINS-48198


vivek:
@reviewbybees

jglick:
Ping me if you need this released and a core PR filed to integrate it.

vivek:
@jglick yes please. Thanks.

PR #140, opened 2018-06-07
promote field injection with declarative constraints

ndeloof:
# Work in progress, do not merge

Plugin developers use to rely on `@DataBoundConstructor`s to get raw data and proceed validation, trim, null check and all that sort of things.

I propose we rely on a declarative approach to define data-binding constraints based on bean validation (jsr 303) annotations. This allows to detect those constraints by reflexion (for sample, to be used by Configuration-as-Code), and as such could also be used to improve the jelly tags library to offer client-side validation.

This is reference implementation for JPE-[TBD]

### DIFF
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -12,4 +12,12 @@
   <artifactId>annotations</artifactId>
 
 
+  <dependencies>
+    <dependency>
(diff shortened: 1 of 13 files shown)


ndeloof:
https://github.com/stapler/stapler/pull/147 do replace this inactive PR

PR #144, opened 2018-07-05
Allow optional st:include with class attribute

daniel-beck:
When writing https://github.com/jenkinsci/login-theme-plugin I stumbled over the following not working (exception thrown) if the specified class did not have the specified page:

    <st:include class="jenkins.model.DefaultSimplePageDecorator" page="simple-footer.jelly" optional="true" />

Tested and confirmed working interactively. Unsure where I'd put tests. Thoughts?

### DIFF
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/IncludeTag.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/IncludeTag.java
@@ -113,6 +113,10 @@ public class IncludeTag extends TagSupport {
         } catch (RuntimeException e) {
             throw e;
         } catch (Exception e) {
+            if(optional) {
+                invokeBody(output);
+                return;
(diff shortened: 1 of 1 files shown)


daniel-beck:
PR builds are fixed by https://github.com/stapler/stapler/pull/145

jelly/src/main/java/org/kohsuke/stapler/jelly/IncludeTag.java
```diff
@@ -113,6 +113,10 @@ public void doTag(XMLOutput output) throws JellyTagException {
         } catch (RuntimeException e) {
             throw e;
         } catch (Exception e) {
+            if(optional) {
```
jglick: What actual exception is thrown? How do you know it is safe to swallow this one?

PR #147, opened 2018-09-24
add bean-validation, @Trim and @DefaultValue annotations

ndeloof:
this PR introduce support for bean-validation (jsr-303) annotations on databound fields, to prevent invalid values being set by data-binding without need for hand-written validation logic

In addition it also introduce two "transformation" annotations:
- `@Trim` to replace need for `Utils.trimToNull` and make constructor / setters simpler
- `@DefaultValue` to let developers define a default value to be used when not set by form (or trimmed to null). Could be used by Jenkins jelly tags to fill default value in web forms

This is the "_make plugin development simpler/safer_" part of [JEP-205](https://github.com/jenkinsci/jep/tree/master/jep/205)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -154,9 +154,15 @@
       <version>1.1.3</version>
     </dependency>
     <dependency>
-      <groupId>org.hibernate</groupId>
+      <groupId>javax.validation</groupId>
+      <artifactId>validation-api</artifactId>
+      <version>2.0.1.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate.validator</groupId>
       <artifactId>hibernate-validator</artifactId>
-      <version>6.0.10.Final</version>
+      <version>6.0.13.Final</version>
+      <scope>runtime</scope>
     </dependency>
 
     <dependency>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -154,6 +154,12 @@
       <version>1.1.3</version>
     </dependency>
     <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-validator</artifactId>
+      <version>6.0.10.Final</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpmime</artifactId>
       <version>4.5</version>
--- a/core/src/main/java/org/kohsuke/stapler/ConstraintsValidationException.java
+++ b/core/src/main/java/org/kohsuke/stapler/ConstraintsValidationException.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2018, Nicolas De Loof
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided
+ * that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright notice, this list of
+ *       conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright notice, this list of
+ *       conditions and the following disclaimer in the documentation and/or other materials
+ *       provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS
+ * OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.kohsuke.stapler;
+
+import javax.validation.ConstraintViolation;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class ConstraintsValidationException extends IllegalArgumentException {
+
+    private Set<ConstraintViolation> violations;
+
+    public ConstraintsValidationException(Set<ConstraintViolation> violations) {
+        super();
+        this.violations = violations;
+    }
+
+    @Override
+    public String getMessage() {
+        return violations.stream().map(v -> v.getPropertyPath() + ": " + v.getMessage()).collect(Collectors.joining(", "));
+    }
+
+    public Set<ConstraintViolation> getViolations() {
+        return violations;
+    }
+}
--- a/core/src/main/java/org/kohsuke/stapler/DefaultValue.java
+++ b/core/src/main/java/org/kohsuke/stapler/DefaultValue.java
@@ -1,0 +1,19 @@
+package org.kohsuke.stapler;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Indicate value to be used when databound payload is missing property <em>or</em> provided value is null
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+@Retention(RUNTIME)
+@Target({FIELD, PARAMETER})
+public @interface DefaultValue {
+
+    String value();
+}
--- a/core/src/main/java/org/kohsuke/stapler/PostProcessor.java
+++ b/core/src/main/java/org/kohsuke/stapler/PostProcessor.java
@@ -1,0 +1,7 @@
+package org.kohsuke.stapler;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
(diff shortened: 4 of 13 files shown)


core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -800,68 +814,80 @@ private Object bindResolve(Object o, JSONObject src) {
...
-                 || !m.getName().startsWith("set")
-                 || m.getParameterTypes().length!=1
-                 || !m.isAnnotationPresent(DataBoundSetter.class))
-                    continue;
+    private <T> void setField(T r, JSONObject j, String name, Field f) {
+        final Object json = getJsonValueOrDefault(j, name, f);
+        if (json == null) return;
+        final Object value = bindJSON(f.getGenericType(), f.getType(), json);
+        try {
+            f.setAccessible(true);
```
jvz: Didn't you yourself publish a library to avoid having to call this directly? Could be useful here.

ndeloof: I indeed experimented with java 9's `VarHandle` but this is not relevant here, `setAccessible` is well supported in java 9+, just reflexion on _internal_ JRE classes is forbidden.
I also try to keep this minimal, to ensure approving this PR won't drive into infinite debates 😇

jvz: Ah ok, sounds good!

ndeloof:
cc @jglick @kohsuke

kohsuke:
Without looking at the actual changes, I love the goal and what this will achieve! Go go go!!

ndeloof:
I'll investigate a PR on Jenkins to get jelly taglib rely on this new @DefaultValue annotation

ndeloof:
@rsandell the idea here is that trim to null should be used with a default value, so that yon don't get nulls. empty string is a very bad substitute for this purpose

ndeloof:
created branch https://github.com/jenkinsci/jenkins/tree/jep-205 to demonstrate use in jenkins-core

ndeloof:
usage demonstrated on https://github.com/jenkinsci/jenkins/pull/3669

ndeloof:
how can we get one step further with this ?

core/pom.xml
```diff
@@ -153,6 +153,12 @@
       <artifactId>jzlib</artifactId>
       <version>1.1.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-validator</artifactId>
+      <version>6.0.10.Final</version>
+    </dependency>
+
```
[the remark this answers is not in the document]
ndeloof: to avoid them the temptation to rely on hibernate-validator internals this would indeed work, but won't prevent classpath conflict if they decide to have another version of hibernate-validator as a dependency

that being said, hibernate-validator internal classes are hardly usable out of validation-api context.

core/src/main/java/org/kohsuke/stapler/ConstraintsValidationException.java
```diff
@@ -0,0 +1,50 @@
...
+package org.kohsuke.stapler;
+
+import javax.validation.ConstraintViolation;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class ConstraintsValidationException extends IllegalArgumentException {
```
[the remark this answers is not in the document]
ndeloof: They would catch ConstraintsValidationException if they want to know about the constraints.
They would catch IllegalArgumentException (or actually, not catch them as an unchecked exception) if they only want to trap random processing errors with payload - which this exception is designed for.

core/pom.xml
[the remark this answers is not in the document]
ndeloof: we actually rely on ParameterMessageInterpolator. Could probably be re-implemented

ndeloof:
`Could not find or load main class org.apache.maven.surefire.booter.ForkedBooter`

could we please get a sane CI service?

jglick:
> could we please get a sane CI service?

Port https://github.com/jenkinsci/pom/pull/34 to fix.

jglick:
(and/or change `Jenkinsfile` to use a more specific tag)

casz:
Wonder what's blocking this from being merged? Would really love to see these changes arrive in Jenkins plugins.

core/src/main/java/org/kohsuke/stapler/PostProcessor.java
```diff
@@ -0,0 +1,7 @@
+package org.kohsuke.stapler;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class PostProcessor {
```
jvz: Is this supposed to be an annotation or have other stuff?

ndeloof: seems to be some draft code I committed by mistake

jeffret-b: @ndeloof, it seems like this draft code should get cleaned up before merging. Is that something you could do?

core/src/main/java/org/kohsuke/stapler/ConstraintsValidationException.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS
+ * OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
Wadeck: For future reviewers, this license header is the "standard" one used in the rest of the library.

core/src/main/java/org/kohsuke/stapler/Replaceable.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public interface Replaceable<T> {
+
+    /**
+     * Replace the data content of current object with new value.
+     * @param newValue
+     * @throws IOException
+     */
+    void replaceBy(T newValue) throws IOException;
+}
```
Wadeck: nit: missing trailing blank line

core/src/main/java/org/kohsuke/stapler/DefaultValue.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Indicate value to be used when databound payload is missing property <em>or</em> provided value is null
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
```
Wadeck: nit: `@since TODO`

core/src/main/java/org/kohsuke/stapler/ConstraintsValidationException.java
```diff
@@ -0,0 +1,50 @@
...
+
+package org.kohsuke.stapler;
+
+import javax.validation.ConstraintViolation;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
```
Wadeck: nit: `@since TODO`

core/pom.xml
```diff
@@ -153,6 +153,18 @@
...
     </dependency>
+    <dependency>
+      <groupId>javax.validation</groupId>
+      <artifactId>validation-api</artifactId>
+      <version>2.0.1.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate.validator</groupId>
+      <artifactId>hibernate-validator</artifactId>
+      <version>6.0.13.Final</version>
```
Wadeck: nit: 6.0.14.Final now (released on Jan 2019)

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -503,6 +525,9 @@ public void bindJSON(Object bean, JSONObject src) {
                 throw (Error)e;
             throw new RuntimeException(x);
         }
+
+        final Set violations = validator.validate(bean);
```
Wadeck: nit: Add generic: `ConstraintViolation<?>` ?

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -328,6 +331,56 @@ public void testFieldInjection() {
...
+        void assertValues() {
+            assertEquals("test",w);
+            assertEquals(1,x);
+            assertEquals("2",y);
+            assertEquals("3",z);
+        }
+    }
+
+    public void testFieldInjectionWithValidation() {
+        BeanValidation r = bind("{x:1,y:'2',z:'3'} }", BeanValidation.class);
```
Wadeck: The json seems malformated due to the two trailing `}`. The test should fail from my PoV.

Actually all the json are invalid, the lib seems to be **really** lenient, it's not just a trailing `,` but a `{}` level issue here.

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -328,6 +331,56 @@ public void testFieldInjection() {
...
+        r.assertValues();
+    }
+
+    public void testFieldInjectionWithValidationFailure() {
+        try {
+            bind("{x:0,y:' ',z:'foo'} }", BeanValidation.class);
```
Wadeck: This test could receive more love in terms of information. As the test above, the JSON seems malformated due to the double `}`, but the test only catch a IAE, without expliciting what is expected.

ndeloof:
I'm closing this PR as I don't plan to invest more time on this, after months waiting for approval. Feel free to cherry-pick code if this topic is some way resurrected.

jglick:
https://github.com/jenkinsci/jenkins/commit/d0b5674b26f0a8e7714fa2a6dac24093b7c0bd27 FTR (in case the stale branch ref is deleted)

PR #148, opened 2018-10-09
Trace Stapler request processing

daniel-beck:
Upstream from https://github.com/jenkinsci/jenkins/pull/3688

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java
+++ b/core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java
@@ -23,9 +23,13 @@
 
 package org.kohsuke.stapler;
 
+import javax.annotation.Nonnull;
 import java.util.List;
 import java.util.ArrayList;
 import java.io.PrintWriter;
+import java.util.ServiceLoader;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import static org.kohsuke.stapler.Stapler.escape;
 
@@ -37,6 +41,8 @@ import static org.kohsuke.stapler.Stapler.escape;
  */
 public class EvaluationTrace {
     private final List<String> traces = new ArrayList<String>();
+
+    private static final Logger LOGGER = Logger.getLogger(EvaluationTrace.class.getName());
 
     public void trace(StaplerResponse rsp, String msg) {
         traces.add(msg);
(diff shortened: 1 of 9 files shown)


daniel-beck:
https://github.com/stapler/stapler/pull/148/commits/3b3105fafb1b44a61bc06fc25bd21e92b5f8ec04 (plus a local `versions:set`) deployed as `1.255-telemetry-1-SNAPSHOT`.

core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java
```diff
@@ -61,4 +67,40 @@ public static EvaluationTrace get(StaplerRequest req) {
...
+            }
+        }
+
+        private static List<ApplicationTracer> tracers;
+
+        @Nonnull
+        private static List<ApplicationTracer> getTracers() {
+            if (tracers == null) {
+                synchronized (ApplicationTracer.class) {
+                    if (tracers == null) {
```
jglick: Double-checked locking. Delete the outer null check.

daniel-beck:
https://github.com/stapler/stapler/pull/148/commits/d33d285f792f15f29242ef0e3616b47771e00851?w=1

core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java
```diff
@@ -61,4 +67,38 @@ public static EvaluationTrace get(StaplerRequest req) {
...
+        protected abstract void record(StaplerRequest req, String message);
+
+        public static void trace(StaplerRequest req, String message) {
+            List<ApplicationTracer> tracers = getTracers();
+            for (ApplicationTracer tracer : tracers) {
+                tracer.record(req, message);
+            }
+        }
+
+        private static volatile List<ApplicationTracer> tracers;
```
jglick: Pointless for this to be `volatile`, since you have the `synchronized` block.

core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java
```diff
@@ -61,4 +67,38 @@ public static EvaluationTrace get(StaplerRequest req) {
...
+                            // robustness
+                            if (LOGGER.isLoggable(Level.FINE)) {
+                                LOGGER.log(Level.FINE, "Exception thrown when notifying tracer", e);
+                            }
+                        }
+                    }
+                    tracers = t;
+                }
+            }
+            return tracers;
```
jglick: This should really be _inside_ the `synchronized` block. I _think_ it is harmless to have it outside, since there will be a synchronization barrier before, but definitely not the best style.

core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java
```diff
@@ -61,4 +67,38 @@ public static EvaluationTrace get(StaplerRequest req) {
...
+            }
+        }
+
+        private static volatile List<ApplicationTracer> tracers;
+
+        @Nonnull
+        private static List<ApplicationTracer> getTracers() {
+            synchronized (ApplicationTracer.class) {
+                if (tracers == null) {
+                    List<ApplicationTracer> t = new ArrayList<>();
```
jglick: You do not need a separate variable. You can initialize `tracers` directly, since it is protected by a synchronization barrier.

jglick:
FTR all you need is something like:

```java
private static List<X> xs;
static synchronized List<X> xs() {
  if (xs == null) {
    xs = new ArrayList<>();
    // populate xs
  }
  return xs;
}
```

Simple and safe.

PR #149, opened 2018-10-10
Actually allow returning null from StaplerProxy

daniel-beck:
It seems Stapler documents returning `null` from `getTarget()` to backtrack (or abort dispatching), but in reality, it doesn't work that way: It's the same as returning `this`.

I noticed this bug when trying to conditionally have an object not handle a request at all (when the user has no permission to `Item/Read` it) and fall back to Stapler failing to route.

Upstream of https://github.com/jenkinsci/jenkins/pull/3690

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Stapler.java
+++ b/core/src/main/java/org/kohsuke/stapler/Stapler.java
@@ -697,9 +697,11 @@ public class Stapler extends HttpServlet {
                 else
                     throw e;    // unprocessed exception
             }
-            if(n==node || n==null) {
+            if (n == node) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -697,9 +697,11 @@ boolean tryInvoke(RequestImpl req, ResponseImpl rsp, Object node ) throws IOExce
                 else
```
jglick: BTW the `Function.renderResponse` behavior appears to be undocumented.

Issue #141, opened 2018-06-08
Downloading a large log file fails in jenkins


diroussel:
I have a jenkins job that has a very large output log:  2,782,890 KB in size.

When I try to download it I get this exception:

```
java.lang.ArithmeticException: The byte count 2846143684 is too large to be converted to an int
	at org.apache.commons.io.output.CountingOutputStream.getCount(CountingOutputStream.java:71)
	at org.kohsuke.stapler.framework.io.LargeText.writeLogTo(LargeText.java:237)
	at hudson.console.AnnotatedLargeText.writeLogTo(AnnotatedLargeText.java:154)
	at org.kohsuke.stapler.framework.io.LargeText.writeLogTo(LargeText.java:194)
	at hudson.console.AnnotatedLargeText.writeLogTo(AnnotatedLargeText.java:145)
	at io.jenkins.blueocean.service.embedded.rest.LogResource.writeLogs(LogResource.java:75)
	at io.jenkins.blueocean.service.embedded.rest.LogResource.writeLog(LogResource.java:51)
	at io.jenkins.blueocean.service.embedded.rest.LogResource.doIndex(LogResource.java:37)
	at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:627)
	at org.kohsuke.stapler.Function$MethodFunction.invoke(Function.java:343)
	at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:184)
	at org.kohsuke.stapler.Function.bindAndInvokeAndServeResponse(Function.java:117)
	at org.kohsuke.stapler.IndexDispatcher.dispatch(IndexDispatcher.java:26)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:715)
Caused: javax.servlet.ServletException
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:765)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:845)
	at org.kohsuke.stapler.MetaClass$3.doDispatch(MetaClass.java:209)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:58)
        ..

```

Looking at CountingOutputStream there is a getByteCount() method that returns a long.  But LargeText is calling getCount() which does a length check, and throws this exception, before casting to an int.

LargeText.writeLogTo() returns a long, so it should be calling getByteCount() and dealing with longs, and thus allow large log files to be downloaded without this exception.

[...]

PR #151, opened 2018-11-07
Enable large log files to be downloaded. Fix for #141

diroussel:
At the end of LargeText.writeLogTo() the size size is returned back as
a long, but the calculation is done with int math, not long.  So we get
a problem for large files.

This fix enables files longer than Integer.MAX_VALUE to be handled.

Unfortunately the only way to test this is by writing out a file that is
larger than Integer.MAX_VALUE.  This test takes 24 seconds to run on my
MacBookPro 15" (2015), which has a very fast SSD.  To test in any other
way would mean much larger changes to LargeFile, just for test purposes.

### DIFF
--- a/core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
@@ -80,6 +80,7 @@ public class LargeTextTest {
 
     @Issue("#141")
     @Test
+    @Ignore
     public void writeLogToWithLargeFile() throws Exception {
         Path path = Files.createTempFile("stapler-test", ".log.gz");
         long size = Integer.MAX_VALUE + 256L;
--- a/core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
@@ -68,6 +68,7 @@ public class LargeTextTest {
         LargeText t;
         try (ByteBuffer bb = new ByteBuffer()) {
             bb.write(text.getBytes(), 0, text.length());
+
             t = new LargeText(bb, true);
         }
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
--- a/core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
@@ -26,11 +26,21 @@
 
 package org.kohsuke.stapler.framework.io;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
 import java.io.ByteArrayOutputStream;
 import java.io.EOFException;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import com.google.common.base.Strings;
+import org.apache.commons.io.output.NullOutputStream;
 import org.junit.Test;
-import static org.junit.Assert.*;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
```diff
@@ -53,13 +63,53 @@ public void writeLogTo() throws Exception {
...
+
+            t = new LargeText(bb, true);
+        }
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         assertEquals(text.length(), t.writeLogTo(start, baos));
         return baos.toString();
     }
 
+    @Issue("#141")
+    @Test
```
jglick: Great but please `@Ignore` it so we do not crash CI.

core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
```diff
@@ -53,13 +63,53 @@ public void writeLogTo() throws Exception {
...
+    public void writeLogToWithLargeFile() throws Exception {
+        Path path = Files.createTempFile("stapler-test", ".log.gz");
+        long size = Integer.MAX_VALUE + 256L;
+        writeLargeFile(path, size);
+
+        LargeText t = new LargeText(path.toFile(), StandardCharsets.US_ASCII, true);
```
jglick: For [JEP-210](https://jenkins.io/jep/210) and friends I have been thinking that we anyway need to make `Source` and `Session` (or some slight variants) `public` and add a `LargeText` constructor taking `Source`. That would allow this test to be written in a way that does not actually consume any disk or RAM.

jglick:
BTW you will need to `git merge master` to fix CI builds.

core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
diroussel: @jglick done. @Ignore added

core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
diroussel: Yes that would be good.  LargeText should be mockable, but the reliance on File gets in the way.

core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
```diff
@@ -26,17 +26,29 @@
...
+import org.junit.Ignore;
 import org.junit.Test;
-import static org.junit.Assert.*;
 import org.jvnet.hudson.test.Issue;
 
 public class LargeTextTest {
 
     @Issue("JENKINS-37664")
     @Test
+    @Ignore
```
jglick: Wait, why does this existing test need to be ignored?

core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
```diff
@@ -53,13 +64,54 @@ public void writeLogTo() throws Exception {
...
+        try (OutputStream stream = new FileOutputStream(path.toFile())) {
+            long remaining = size;
+            while (remaining > chunkSize) {
+                stream.write(bytesChunk);
+                remaining -= chunkSize;
+            }
+            stream.write(bytesChunk, 0, (int) remaining);
+            stream.flush();
+        } catch (IOException e) {
+            fail(e.getMessage());
```
jglick: Note that you can just let the method throw up the exception and the test will fail just the same.

core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
diroussel: ah, sorry, my mistake. fixed

Issue #153, opened 2018-12-05
security question re: recent Jenkins advisory


attritionorg:
Regarding the recent Jenkins advisory: https://jenkins.io/security/advisory/2018-12-05/

"_Jenkins uses the Stapler web framework for HTTP request handling. Stapler’s basic premise is that it uses reflective access to code elements matching its naming conventions. For example, any public method whose name starts with get, and that has a String, int, long, or no argument can be invoked this way on objects that are reachable through these means. As these naming conventions closely match common code patterns in Java, accessing crafted URLs could invoke methods never intended to be invoked this way._"

Is this a fundamental issue in Stapler, or is this specific to Jenkins implementation? Curious if other applications should be concerned about this. Thanks!

rschultheis:
I came here with the same question.  This appears to be related to this commit: 
https://github.com/stapler/stapler/commit/28e8eba822a0df9dcd64d20eb63d8ab5f6ee2980
It would seem there is a vulnerability in stapler that will be fixed in 1.256 once it is published.  Right now the latest version in Maven is 1.255.

Perhaps @daniel-beck or @Wadeck can add some info as the authors of that commit?

daniel-beck:
> Is this a fundamental issue in Stapler, or is this specific to Jenkins implementation? Curious if other applications should be concerned about this. Thanks!

This is actually a good question, and I'm not quite sure of the answer myself.

It's a fundamental property in Stapler's use of naming convention for its routing, and carefully written applications can just not expose public fields or methods matching the conventional `getWhatever` name on objects they make reachable through routing unless they're intended to be routed, and thereby largely sidestep the problem, with the single exception of `Object#getClass()` and  anything reachable transitively through that. So it really depends on the application whether it uses these (very common) code patterns that would allow this to be exploited. This is the case in Jenkins, and it needed additional protections in the form of implementations of the newly added SPI.

If you're using Stapler as your framework, I recommend setting the `org.kohsuke.stapler.Dispatcher.TRACE` property to `true`, and look at HTTP response headers, as well as the output of the 404 error page for URLs that don't exist to understand what is considered routable by Stapler, and follow that transitively to interesting places to understand whether your application is impacted.

FWIW we searched a while back for users of Stapler as a web framework and came up empty, and between this and not getting any responses to other Stapler-related security fixes we just assumed there was nobody else using it (besides Hudson, but that project appeared dormant last I checked).

daniel-beck:
[...]

PR #154, opened 2018-12-11
Always reject the #getClass() method

daniel-beck:
There seems to be no legitimate use case for allowing `#getClass()` to be called, and not allowing it will make the following statement true:

> It's a fundamental property in Stapler's use of naming convention for its routing, and carefully written applications can just not expose public fields or methods matching the conventional `getWhatever` name on objects they make reachable through routing unless they're intended to be routed, and thereby sidestep the problem.

https://github.com/stapler/stapler/issues/153#issuecomment-445966538 has a more limited variant of that, and there's no reason it should not be true.

Untested.

@Wadeck

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/MetaClass.java
+++ b/core/src/main/java/org/kohsuke/stapler/MetaClass.java
@@ -182,7 +182,7 @@ public class MetaClass extends TearOffSupport {
             });
         }
 
-        FunctionList getMethods = node.methods.prefix("get");
+        FunctionList getMethods = node.methods.prefix("get").filter(m -> !m.getSignature().equals("method java.lang.Object getClass"));
         FunctionList filteredGetMethods;
         if(LEGACY_GETTER_MODE || webApp.getFilterForGetMethods() == null){
             LOGGER.log(Level.FINE, "Stapler is using the legacy GETTER_MODE");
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/MetaClass.java
```diff
@@ -182,7 +182,7 @@ public String toString() {
             });
         }
 
-        FunctionList getMethods = node.methods.prefix("get");
+        FunctionList getMethods = node.methods.prefix("get").filter(m -> !m.getSignature().equals("method getClass"));
```
daniel-beck: `method java.lang.Object getClass` actually.

So it looks like this deserves a test after all.

core/src/test/java/org/kohsuke/stapler/MetaClassTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class MetaClassTest extends TestCase {
+    public void testGetObjectProhibited() throws Exception {
+        MetaClass metaClass = new MetaClass(new WebApp(new MockServletContext()), Klass.java(Object.class));
+
+        // ensure no getClass dispatcher for Object
+        Assert.assertFalse(metaClass.dispatchers.stream().filter(d -> d instanceof NameBasedDispatcher).anyMatch(d -> ((NameBasedDispatcher) d).name.equals("class")));
+
+        // in fact, there should be no name based dispatchers at all
+        Assert.assertFalse(metaClass.dispatchers.stream().anyMatch(d -> d instanceof NameBasedDispatcher));
```
jglick: Could just drop the first assertion in that case?

jglick:
merge with `master` to get CI

daniel-beck:
> merge with `master` to get CI

Improved the code instead 🤓

PR #155, opened 2018-12-14
Added DispatcherTest.testProtectedMethodDispatch

jglick:
Clarifies behavior of aca8c023adbc051c80a5e1d785ca76411ff9ed32 based on [this check](https://github.com/stapler/stapler/blob/cebe82d8aee82f93797f315a230fcc74ff950f64/core/src/main/java/org/kohsuke/stapler/ClassDescriptor.java#L168).

### DIFF
--- a/core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/DispatcherTest.java
@@ -366,6 +366,26 @@ public class DispatcherTest extends JettyTestCase {
         }
     }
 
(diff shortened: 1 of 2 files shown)


jvz:
I will note that I prefer using `ExpectedException` from JUnit, though if you don't have a nice Hamcrest matcher for your exceptions, then it might be more trouble than it's worth.

jglick:
`ExpectedException` looks like a lot more trouble than it is worth, especially since you can only use it once per test.

PR #156, opened 2019-02-05
[JEP-216] Add SPI for localized resource lookup

daniel-beck:
Upstream of https://github.com/jenkinsci/jenkins/pull/3729 etc.


core/src/main/java/org/kohsuke/stapler/LocaleDrivenResourceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public abstract class LocaleDrivenResourceProvider {
+    /**
+     * Returns the URL corresponding to the specified resource path.
+     *
+     * The path can take two forms:
+     *
+     * <ul>
+     *     <li>A full URL represented as string. In Jenkins this is typically plugin webapp resources.</li>
+     *     <li>A path with leading slash. These are actual webapp resources, with that being the implicit base directory.</li>
```
daniel-beck: A potentially better design would be to have independent methods `lookupUrl` and `lookupPath`. Thoughts?

LinuxSuRen: Can't figure out the use cases. The interface method may not need to take care of those details.

core/src/main/java/org/kohsuke/stapler/LocaleDrivenResourceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return the URL to the file, if it is to be overridden, null otherwise.
+     */
+    @CheckForNull
+    abstract public URL lookup(@Nonnull String path);
+
+    private static List<LocaleDrivenResourceProvider> localeDrivenResourceProviders;
+
+    private static final Logger LOGGER = Logger.getLogger(LocaleDrivenResourceProvider.class.getName());
+
+    private static synchronized List<LocaleDrivenResourceProvider> getLocaleDrivenResourceProviders() {
```
jvz: You can limit the synchronization to after you check for null (but then check for null again). Make the field volatile and you've got a threadsafe lazy initialized value.

daniel-beck: I'm going with @jglick's verbatim advice from an earlier equivalent change here, and not touching this 😃

core/src/main/java/org/kohsuke/stapler/LocaleDrivenResourceProvider.java
jeffret-b: I'm generally in favor of using explicitly named separate methods when there is a clear separation of responsibility or implementation. In this case, though, I'm not sure that separation exists. Both forms are valid types of URLs.

Is there some reason we think the caller or implementor might need to handle these two forms differently or separately?

daniel-beck: > Is there some reason we think the caller or implementor might need to handle these two forms differently or separately?

It's most of what https://github.com/daniel-beck/localization-support-plugin/blob/master/src/main/java/io/jenkins/plugins/localization/support/stapler/LocaleDrivenResourceProviderImpl.java does.

"From a plugin or not" is the same distinction as the two variants documented here.

daniel-beck: > Both forms are valid types of URLs.

Note that the second one (plugin resources) is just `/path/on/file/system` pointing to `$JENKINS_HOME/plugins/<ID>/…`.

core/src/main/java/org/kohsuke/stapler/LocaleDrivenResourceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param path the path to the resource
+     * @return the URL to the file, if it is to be overridden, null otherwise.
+     */
+    @CheckForNull
+    abstract public URL lookup(@Nonnull String path);
+
+    private static List<LocaleDrivenResourceProvider> localeDrivenResourceProviders;
+
+    private static final Logger LOGGER = Logger.getLogger(LocaleDrivenResourceProvider.class.getName());
```
Wadeck: nit: field declarations should be done before methods declarations

core/src/main/java/org/kohsuke/stapler/LocaleDrivenResourceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.ServiceLoader;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Service provider interface allowing to hook into webapp resource lookup.
+ *
+ * This cannot be made a property of WebApp as other behavior customizations, as webapp resource lookup is done before we have StaplerRequest/StaplerResponse.
+ */
+public abstract class LocaleDrivenResourceProvider {
```
Wadeck: Add the `implements ExtensionPoint` ?

daniel-beck: Can't, that's in Jenkins.

daniel-beck: See also https://github.com/jenkinsci/jenkins/pull/3729/files#diff-ca4047332ddf7ab84f19e41ee830640eR40

core/src/main/java/org/kohsuke/stapler/LocaleDrivenResourceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+    private static List<LocaleDrivenResourceProvider> localeDrivenResourceProviders;
+
+    private static final Logger LOGGER = Logger.getLogger(LocaleDrivenResourceProvider.class.getName());
+
+    private static synchronized List<LocaleDrivenResourceProvider> getLocaleDrivenResourceProviders() {
+        if (localeDrivenResourceProviders == null) {
+            localeDrivenResourceProviders = new ArrayList<>();
+            for (LocaleDrivenResourceProvider provider : ServiceLoader.load(LocaleDrivenResourceProvider.class)) {
+                localeDrivenResourceProviders.add(provider);
+                LOGGER.log(Level.INFO, "Registered LocaleDrivenResourceProvider: " + provider);
```
jglick: Probably could be made `FINE` since this is not exactly interesting news to an admin to be shown on every startup.

PR #157, opened 2019-02-19
Add incrementals

jeffret-b:
I'd like to incrementalize stapler. This will be handy in general and will make it easier to do some other changes we want.

Since stapler isn't a plugin, I couldn't follow the easy mechanism. Instead I followed the step-by-step, manual instructions. The changes match what I've seen the automated, easy version do, plus the addition of the profiles needed because the parent is different.

While I've used this to effectively generate incrementals locally I don't know that it is sufficient to automatically generate them in Jenkins CI environment. Specifically, updates my be required for the stapler settings in repository-permissions-updater.

Do we want a Jira ticket associated with this change?

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-2</version>
+    <version>1.0-beta-7</version>
   </extension>
 </extensions>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools.incrementals</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
+    <version>1.0-beta-2</version>
+  </extension>
+</extensions>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <groupId>org.kohsuke.stapler</groupId>
   <artifactId>stapler-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.257-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
 
   <name>Stapler</name>
   <description>Stapler HTTP request handling engine</description>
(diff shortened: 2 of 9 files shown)


.mvn/extensions.xml
```diff
@@ -0,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools.incrementals</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
+    <version>1.0-beta-2</version>
```
jvz: Can use the latest version (I think beta-7)

pom.xml
```diff
@@ -281,4 +285,141 @@
...
+                    </rule>
+                  </rules>
+                </configuration>
+              </execution>
+            </executions>
+            <dependencies>
+              <dependency>
+                <groupId>io.jenkins.tools.incrementals</groupId>
+                <artifactId>incrementals-enforcer-rules</artifactId>
+                <version>1.0-beta-4</version>
```
jvz: I think this might also be out of date based on what we've done in audit-log-plugin.

jeffret-b:
Good catch. I just copied from the documentation and somewhere else, which are clearly out of date. I didn't check for latest version.

pom.xml
```diff
@@ -37,7 +37,7 @@
     <connection>scm:git:git://github.com/stapler/stapler.git</connection>
     <developerConnection>scm:git:ssh://git@github.com/stapler/stapler.git</developerConnection>
     <url>https://github.com/stapler/stapler</url>
-    <tag>HEAD</tag>
+    <tag>${scmTag}</tag>
```
varyvol: Does this work fine with creating releases?

jeffret-b: Apparently it does. Here's the documentation on it: [Incrementals](https://www.google.com/url?q=https://rmh.psdschools.org/webfm/1875&sa=U&ved=0ahUKEwjpxcvIo9fgAhVFYK0KHdbyAN0QFggEMAA&client=internal-uds-cse&cx=004191848107721260813:sww1jghnesa&usg=AOvVaw0zpO40o24JrwsU9n9fkD1t)

And the changes the automated incrementalization tool added in Remoting for reference: [Add incremtentals to Remoting](https://github.com/jenkinsci/remoting/commit/33bc3f99f7155ec090e01bbd158cd3105cd2e487). I've successfully released Remoting after that.

jvz: scmTag is defined in the parent pom typically, and the release stuff sets that automatically (from incrementals).

jeffret-b:
Fixed the link. Don't use the quoted one. Not sure how that got there.

daniel-beck:
IIRC @jglick pointed out this wouldn't work.

jeffret-b:
It won't fully work, because it's not fully connected to the Jenkins CI system. The automatic publishing of incrementals won't work. I think it could still be useful for development, though.

jglick:
> The automatic publishing of incrementals won't work. I think it could still be useful for development, though.

Yes you should be able to

```sh
mvn -Dset.changelist clean deploy
```

if you have permission to the repository.

daniel-beck:
> it's not fully connected to the Jenkins CI system

There's https://ci.jenkins.io/job/Stapler/job/stapler/ though, or what do you mean?

jglick:
Yes there is, but actually deployment requires that an Azure function be called, and that function would in turn check `repository-permission-updater` stuff which is not going to work since this repo normally gets deployed to OSSRH not the Jenkins Artifactory.

PR #161, opened 2019-04-19
Make LocaleDrivenResourceProvider quieter

jglick:
Was irritated to see an extra message in the Jenkins startup log. https://github.com/stapler/stapler/pull/156#discussion_r273059383


daniel-beck:
Not sure about this one, given how extensive the potential impact is.

PR #164, opened 2019-06-27
[JENKINS-53322] Use standard dom4j distribution.

jeffret-b:
See [JENKINS-53322](https://issues.jenkins-ci.org/browse/JENKINS-53322).

I've been working on eliminating the need for the Jenkins-specific fork of dom4j and adopting the latest, standard release, 2.1.1. There are two aspects 1) dealing with the changes in the fork, which are minor and apparently not consequential, and 2) upgrading a couple of places to match changes in the dom4j API since long ago.

I've run a lot of builds and tests on various components to check these changes. I have found a few places that require changes, A) here in Stapler, B), a test in Jenkins core for API changes, C) licensing packaging change in Jenkins, D) and a couple of minor changes in ec2-plugin for API changes. My PR for ec2-plugin jenkinsci/ec2-plugin#372 works on both the old and new API and has already been merged in. See jenkinsci/jenkins#4089 to make B) and C) behave for both old and new dom4j.

The dom4j changes in the Jenkins don't make any sense. There is another way to do about the same thing, as demonstrated by this PR. The object model makes more sense if the appropriate class is chosen rather than trying to make the one class behave more like the other. Significantly, I can't find any evidence that the modification is ever actually used. There are no tests in the dom4j fork to validate it. The Stapler tests, limited as they are, pass with these changes and do not run through the paths introduced in the Jenkins dom4j fork. The same is true for tests within the Jenkins project and the UI testing I've performed. Where possible, I've compiled and testing plugins or libraries with dependencies or references to dom4j and found no further problems.

This PR cannot be included in Jenkins (via an updated Stapler) version until jenkinsci/jenkins#4089 to resolve B) and C) is merged.

### DIFF
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/DefaultScriptInvoker.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/DefaultScriptInvoker.java
@@ -24,14 +24,14 @@
 package org.kohsuke.stapler.jelly;
 
 import org.apache.commons.jelly.JellyContext;
-import org.apache.commons.jelly.JellyTagException;
-import org.apache.commons.jelly.Script;
-import org.apache.commons.jelly.XMLOutput;
-import org.apache.commons.jelly.XMLOutputFactory;
-import org.apache.commons.jelly.impl.TagScript;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
+import org.apache.commons.jelly.Script;
+import org.apache.commons.jelly.JellyTagException;
+import org.apache.commons.jelly.XMLOutput;
+import org.apache.commons.jelly.XMLOutputFactory;
+import org.apache.commons.jelly.impl.TagScript;
 
 import javax.annotation.Nonnull;
 import javax.servlet.ServletContext;
@@ -65,11 +65,15 @@ public class DefaultScriptInvoker implements ScriptInvoker, XMLOutputFactory {
 
     protected XMLOutput createXMLOutput(StaplerRequest req, StaplerResponse rsp, Script script, Object it) throws IOException {
         // TODO: make XMLOutput auto-close OutputStream to avoid leak
-        HTMLWriterOutput hwo = HTMLWriterOutput.create(createOutputStream(req, rsp, script, it));
         String ct = rsp.getContentType();
-        if (ct != null && !ct.startsWith("text/html"))
-            hwo.useHTML(false);
-        return hwo;
+        XMLOutput output;
+        if (ct != null && !ct.startsWith("text/html")) {
+            output = XMLOutput.createXMLOutput(createOutputStream(req, rsp, script, it));
+        } else {
+            output = HTMLWriterOutput.create(createOutputStream(req, rsp, script, it));
+
+        }
+        return output;
     }
 
     private boolean doCompression(Script script) {
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
@@ -76,6 +76,6 @@ public class HTMLWriterOutput extends XMLOutput {
 
     @Deprecated
     public void useHTML(boolean enabled) {
-        LOGGER.fine("Deprecated: HTMLWriterOutput.useHtml() has been deprecated and its functionality removed.");
+        LOGGER.log(Level. WARNING, null, new UnsupportedOperationException("Deprecated: HTMLWriterOutput.useHtml() has been deprecated and its functionality removed."));
     }
 }
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
@@ -64,18 +64,11 @@ public class HTMLWriterOutput extends XMLOutput {
         hw.setEscapeText(escapeText);
         this.htmlWriter = hw;
         this.format = fmt;
+        format.setExpandEmptyElements(true);
     }
 
     @Override public void close() throws IOException {
         htmlWriter.close();
     }
 
-    /**
-     * False to turn off HTML handling and reenable {@code />} for any empty XML element.
-     * True to switch back to default mode with HTML handling.
-     */
-    public void useHTML(boolean enabled) {
-        htmlWriter.setEnabled(enabled);
-        format.setExpandEmptyElements(enabled);
-    }
 }
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
+import java.util.logging.Logger;
 
 /**
  * Wrapper for XMLOutput using HTMLWriter that can turn off its HTML handling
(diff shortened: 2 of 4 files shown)


jeffret-b:
Projects in the jenkinsci organization that reference dom4j in the pom. I built and tested each of these with updated versions.

[grep.txt](https://github.com/stapler/stapler/files/3336598/grep.txt)

jeffret-b:
Results of search through projects in the jenkinsci organization for mention of "dom4j" in Java files. This search found the API compilation issue in ec2-plugin but nothing further interesting.

[grep2.txt](https://github.com/stapler/stapler/files/3336605/grep2.txt)

jeffret-b:
Results of search through project in jenkinsci organization for "selectNodes", which could relate to a dom4j API change in moving to 2.1.1. This found no additional interesting results.
[grep3.txt](https://github.com/stapler/stapler/files/3336615/grep3.txt)

jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
```diff
@@ -64,18 +64,11 @@ private HTMLWriterOutput(HTMLWriter hw, OutputFormat fmt, boolean escapeText) {
...
 
     @Override public void close() throws IOException {
         htmlWriter.close();
     }
 
-    /**
-     * False to turn off HTML handling and reenable {@code />} for any empty XML element.
-     * True to switch back to default mode with HTML handling.
-     */
-    public void useHTML(boolean enabled) {
```
oleg-nenashev: Can be reverted to avoid the major release

MRamonLeon: Maybe you can set it as @Deprecated and print something in logs to make it clear that this method is not going to be taking into account when rendering xml.

jeffret-b: Great idea. It was very useful to remove it as I was building and searching for any uses so there would be a hard failure. Now that's done, it's good to put it back into prevent anything out there that I may not have caught.

jelly/src/main/java/org/kohsuke/stapler/jelly/DefaultScriptInvoker.java
```diff
@@ -65,11 +66,15 @@ public void invokeScript(StaplerRequest req, StaplerResponse rsp, Script script,
 
     protected XMLOutput createXMLOutput(StaplerRequest req, StaplerResponse rsp, Script script, Object it) throws IOException {
         // TODO: make XMLOutput auto-close OutputStream to avoid leak
-        HTMLWriterOutput hwo = HTMLWriterOutput.create(createOutputStream(req, rsp, script, it));
         String ct = rsp.getContentType();
-        if (ct != null && !ct.startsWith("text/html"))
-            hwo.useHTML(false);
-        return hwo;
+        XMLOutput output;
```
jvz: I'd be careful about merge conflicts here as this method was refactored in the security update.

jeffret-b:
Try rebuilding.

jelly/src/main/java/org/kohsuke/stapler/jelly/DefaultScriptInvoker.java
jeffret-b: Hmmm ... looks like that hasn't been merged in yet.

Wadeck: https://github.com/stapler/stapler/pull/166

jeffret-b: Looks like this method wasn't touched in the recent security update, but other unrelated methods in the file were.

jeffret-b:
Any specific tooling, automated tests, or other libraries come to mind?

I tested the ones I could find that have some connection to dom4j, particularly our version of. The JTH library is one that I tested. In that case as long as the stapler versions used in the JTH and in the Jenkins under test match, it worked fine. Just don't try to reference a different stapler in JTH than Jenkins uses.

Wadeck:
@jeffret-b and did you try PCT and ATH with the modified Stapler version?

jeffret-b:
kick off rebuild

jeffret-b:
Kick again. Looks like a flaky test.

oleg-nenashev:
@jglick would you be able to review and release it if this is ready?

jeffret-b:
There are a couple of test failures that showed up on the CI run of Jenkins yesterday. I didn't see them when I ran jenkins/test locally before. I'm looking into them.

As suggested by Daniel and Wadeck, I'm also trying to get runs of PCT and ATH.

jeffret-b:
Put this on hold for now until I can figure out what's going on with the new test failure.

jeffret-b:
I've figured out basically what's going on with the new test failure. This new test was introduced with recent work on SECURITY-534. This test fails only for the url "annotated/root": https://github.com/jenkinsci/jenkins/blob/master/test/src/test/java/jenkins/security/stapler/StaplerDispatchValidatorTest.java#L78 . The significant difference between this scenario and others is that [`Annotated`](https://github.com/jenkinsci/jenkins/blob/master/test/src/test/java/jenkins/security/stapler/StaplerDispatchValidatorTest.java#L169) uses [`l:layout`](https://github.com/jenkinsci/jenkins/blob/master/test/src/test/resources/jenkins/security/stapler/StaplerDispatchValidatorTest/Annotated/root.jelly#L26). 

The test failure is that it expects the `asText()` to return "Fragment" but with dom4j 2.1.1, it returns "FragmentSkip to content". This is because this [`<a href>`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/layout/layout.jelly#L180) in `layout.jelly` gets included inside the first paragraph with the id "frag". With dom4j 1.6.1, that href exists outside that paragraph tag. The new test checks for all text within an element, including children, so it gets that merged text.

I have checked two different approaches to resolve the test failure. 
1. Change the test to be more precise so that it only checks the `nodeValue` and doesn't pick up child text. (`assertEquals("Fragment", root.getElementById("frag").getChildNodes().get(0).getNodeValue());`)
2. Wrap `layout.jelly:180` in its own paragraph tag.

Significantly, I haven't been able to detect any real differences between what dom4j 1.6.1 and 2.1.1 produce. The UI looks the same and I haven't detected any behavior differences. The only way to detect the difference is by examining the page source. I think that the 2.1.1 are more correct.

jeffret-b:
I'm still working on running the PCT and ATH to see if they find anything else interesting.

jvz:
The test was written simply to assert that it can see the contents and verify it was included. Whatever method is used to check this is completely fine to relax to some form of containsString() type of thing.

jeffret-b:
I added my proposed fix to make the assertion in StaplerDispatchValidatorTest#canViewRoot more precise into my WIP PR for Jenkins to use this new Stapler version, jenkinsci/jenkins#4130. This revised check works for current versions and with the standard dom4j.

I finally managed to get ATH and PCT runs. They found nothing new, only pre-existing failures.

jeffret-b:
Tests are passing. Looks like everything is good to go again on this one.

jeffret-b:
I also tried some dependency searches with jenkins-infra/usage-in-plugins. It didn't find much new. The only notable one was read-only-configurations-plugin. I did a slightly different search to find its usage of the related stapler methods. Adding back in `useHTML()`, as I did a while back, eliminated any problem I've been able to find with that plugin. I couldn't quite get the plugin to work, though I didn't spend a long time on it. Reading the global configuration worked fine before and after. I wasn't able to get it to show the job configuration. An old thread on the Jenkins user email list said some configuration had to be adjusted to make it work but I couldn't find that element.

The rest of the hits were ones I had already found or irrelevant. I was able to craft better, more thorough or more precise, searches using grep.

jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
```diff
@@ -64,18 +67,15 @@ private HTMLWriterOutput(HTMLWriter hw, OutputFormat fmt, boolean escapeText) {
...
 
-    /**
-     * False to turn off HTML handling and reenable {@code />} for any empty XML element.
-     * True to switch back to default mode with HTML handling.
-     */
+    @Deprecated
     public void useHTML(boolean enabled) {
-        htmlWriter.setEnabled(enabled);
-        format.setExpandEmptyElements(enabled);
+        LOGGER.fine("Deprecated: HTMLWriterOutput.useHtml() has been deprecated and its functionality removed.");
```
jglick:
```suggestion
        LOGGER.log(Level. WARNING, null, new UnsupportedOperationException("Deprecated: HTMLWriterOutput.useHtml() has been deprecated and its functionality removed."));
```

since no one is going to see this message otherwise.

jelly/src/main/java/org/kohsuke/stapler/jelly/HTMLWriterOutput.java
```diff
@@ -64,18 +67,15 @@ private HTMLWriterOutput(HTMLWriter hw, OutputFormat fmt, boolean escapeText) {
...
 
-    /**
-     * False to turn off HTML handling and reenable {@code />} for any empty XML element.
-     * True to switch back to default mode with HTML handling.
-     */
+    @Deprecated
     public void useHTML(boolean enabled) {
-        htmlWriter.setEnabled(enabled);
-        format.setExpandEmptyElements(enabled);
+        LOGGER.log(Level. WARNING, null, new UnsupportedOperationException("Deprecated: HTMLWriterOutput.useHtml() has been deprecated and its functionality removed."));
```
jglick: needs to be imported (compilation error)

PR #165, opened 2019-06-28
A little bit of pom cleanup.

jeffret-b:
Primarily about using https.

### DIFF
--- a/jrebel/pom.xml
+++ b/jrebel/pom.xml
@@ -35,19 +35,4 @@
     </dependency>
   </dependencies>
 
-  <!--
-  <repositories>
-    <repository>
-      <id>zt-repo</id>
-      <url>http://repos.zeroturnaround.com/nexus/content/groups/zt-public/</url>
-      <snapshots>
-        <enabled>false</enabled>
-      </snapshots>
-      <releases>
-        <enabled>true</enabled>
-        <checksumPolicy>ignore</checksumPolicy>
-      </releases>
-    </repository>
-  </repositories>
-  -->
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -24,13 +24,13 @@
     <module>jrebel</module>
   </modules>
 
-  <url>http://stapler.kohsuke.org/</url>
+  <url>https://stapler.kohsuke.org/</url>
 
   <licenses>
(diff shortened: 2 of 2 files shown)


jrebel/pom.xml
```diff
@@ -35,19 +35,4 @@
     </dependency>
   </dependencies>
 
-  <!--
```
jeffret-b: This one isn't being used (commented out). The URL doesn't exist. And is http. Seems best to just remove this section.

pom.xml
```diff
@@ -67,14 +67,14 @@
   <repositories>
     <repository>
       <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
+      <url>https://repo.jenkins-ci.org/public/</url>
```
jeffret-b: These don't work at either http or https but it's better convert them to https anyway.

jeffret-b:
@daniel-beck @Wadeck @jvz

pom.xml
Wadeck: WDYM? they are used by the publisher plugin, not meant to be accessible directly in the browser.

daniel-beck: Browsing at the virtual repo base URL is prohibited for performance reasons, e.g. https://repo.jenkins-ci.org/public/org/ works.

jeffret-b: Makes sense. Thanks.

pom.xml
```diff
@@ -24,13 +24,13 @@
...
 
-  <url>http://stapler.kohsuke.org/</url>
+  <url>https://stapler.kohsuke.org/</url>
 
   <licenses>
     <license>
       <name>2-clause BSD license</name>
       <distribution>repo</distribution>
-      <url>http://opensource.org/licenses/bsd-license.php</url>
+      <url>https://opensource.org/licenses/bsd-license</url>
```
jglick:
```suggestion
      <url>https://opensource.org/licenses/BSD-2-Clause</url>
```

seems to be canonical, or is at least what is linked to from https://opensource.org/licenses.

jeffret-b: I've seen the first one a number of places, but the second should be goot.

PR #169, opened 2019-08-19
Enable Release Drafter

jglick:
CC @jeffret-b


jglick:
Does not seem to be working, alas.

PR #171, opened 2019-09-22
[JENKINS-41729] Accept xml property files for localization

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-41729

It's difficult to review localization pull requests for Jenkins plugins in github, as property files are encoded.
I have to review them in IDE.

This change makes stapler accept XML files. XML files can be written with UTF-8 and no need to be encoded.
The XML file property feature are available since Java1.5: https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#loadFromXML-java.io.InputStream-


timja:
I also wonder whether we could convert to using Utf-8 in plugins / core as well? 

like https://github.com/kohsuke/localizer/pull/13 but with an opt in approach, we can script updating core, and update the archetype or plugin Pom to opt in by default?

timja:
Closing in favour of https://github.com/jenkinsci/stapler/pull/357

PR #174, opened 2019-10-28
Upgrade to latest commons-beanutils.

jeffret-b:
Pick up the latest improvements and hardenings.

Jenkins appears to package the version referenced in jenkins/core. Maybe this one, but probably not. It would be good to go ahead and update this one for the next stapler release.


jeffret-b:
This turned out to be more difficult than originally believed. The changes I proposed are insufficient and not necessarily the way I want to proceed. We should consider the technical debt of our Jelly release before proceeding in that direction.

PR #176, opened 2019-12-06
Add a mechanism to suppress stack trace.

jeffret-b:
Add a mechanism to control whether `HttpResponses.error()` shows the error details, particularly the stack trace. By default it doesn't show the stack trace, but the interface can be enabled to show it.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
+++ b/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
@@ -23,11 +23,13 @@
 
 package org.kohsuke.stapler;
 
+import javax.annotation.CheckForNull;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.net.URL;
+import java.util.ServiceLoader;
 
 import static javax.servlet.http.HttpServletResponse.SC_MOVED_TEMPORARILY;
 
@@ -37,6 +39,11 @@ import static javax.servlet.http.HttpServletResponse.SC_MOVED_TEMPORARILY;
  * @author Kohsuke Kawaguchi
  */
 public class HttpResponses {
+
+    public interface ErrorCustomizer {
+        @CheckForNull
+        HttpResponseException handleError(int code, Throwable cause);
+    }
 
     public static abstract class HttpResponseException extends RuntimeException implements HttpResponse {
         public HttpResponseException() {
@@ -89,8 +96,15 @@ public class HttpResponses {
     }
 
     public static HttpResponseException error(final int code, final Throwable cause) {
-        if (Boolean.getBoolean(HttpResponses.class.getName() + ".SHOW_STACK_TRACE")) {
-            return new HttpResponseException(cause) {
+        HttpResponseException responseException = null;
+        for (ErrorCustomizer handler : ServiceLoader.load(ErrorCustomizer.class)) {
+            responseException = handler.handleError(code, cause);
+            if (responseException != null) {
+                break;
+            }
+        }
+        if (responseException == null) {
+            responseException = new HttpResponseException(cause) {
                 public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
                     rsp.setStatus(code);
 
@@ -101,7 +115,7 @@ public class HttpResponses {
                 }
             };
         }
-        return errorWithoutStack(code, cause.getMessage());
+        return responseException;
     }
 
     /**
--- a/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
+++ b/core/src/main/java/org/kohsuke/stapler/HttpResponses.java
@@ -37,6 +37,17 @@ import static javax.servlet.http.HttpServletResponse.SC_MOVED_TEMPORARILY;
  * @author Kohsuke Kawaguchi
  */
 public class HttpResponses {
+
+    public interface ErrorDetailsFilter {
+        boolean showStack(final int code, final Throwable cause);
+    }
+
+    private static ErrorDetailsFilter errorDetailsFilter = (code, cause) -> true;
+
+    public static void setErrorDetailsFilter(ErrorDetailsFilter filter) {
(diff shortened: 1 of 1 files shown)


jeffret-b:
@Wadeck @daniel-beck

jeffret-b:
Let's see if incrementals are working yet.

jeffret-b:
Try kicking off incrementals again.

daniel-beck:
I wonder whether it would be nicer to use a similar pattern as I used for the RequirePOST error page that lets the application render a view: https://github.com/stapler/stapler/blob/7d5bf955f3f0d174e760c522359ad8321c946f97/core/src/main/java/org/kohsuke/stapler/interceptor/RequirePOST.java#L40-L63

Not sure we're in the correct level here to do this, though. That way, even Stapler's error pages (and thereby the error pages served by the application through `HttpResponses#error`) would be nicer looking.

As this extends the API of Stapler, bringing this idea up now instead of considering this just a future enhancement. Since nobody else uses Stapler, it would be easy enough to defer though.

jeffret-b:
I adopted Daniel's suggestion of tying the implementation back to Jenkins again, this time using the oops page. I was concerned that we would have things in the proper space and stack but it seems to work fine.

Please review @jvz @daniel-beck @Wadeck

core/src/main/java/org/kohsuke/stapler/HttpResponses.java
```diff
@@ -37,6 +39,12 @@
  * @author Kohsuke Kawaguchi
  */
 public class HttpResponses {
+
+    public interface ErrorCustomizer {
+        @CheckForNull
```
jvz: Nit: being a new API, it might be preferable to use `Optional<HttpResponseException>` instead, though thanks to the pervasive tooling, the annotation works almost the same, so up to you.

Wadeck: Using new code style / pattern on an existing code base makes the refactoring harder and the new stuff / trend will change every x years. I am against such changes that are reducing the consistency of the code style of the project. As you say, there is no added value.

core/src/main/java/org/kohsuke/stapler/HttpResponses.java
```diff
@@ -88,16 +96,26 @@ public static HttpResponseException error(Throwable cause) {
...
-
-                rsp.setContentType("text/plain;charset=UTF-8");
-                PrintWriter w = new PrintWriter(rsp.getWriter());
-                cause.printStackTrace(w);
-                w.close();
+        HttpResponseException responseException = null;
+        for (ErrorCustomizer handler : ServiceLoader.load(ErrorCustomizer.class)) {
+            responseException = handler.handleError(code, cause);
+            if (responseException != null) {
+                break;
```
Wadeck: ❌ Any reason to not follow the existing pattern from Daniel? I think especially about the "first to handle" and the classload parameter in `.load()`.
If there are problems the other code should be changed and both using the same approach. So that we can prevent discrepancies.

`break` => For me the break is a no-no but actually inside Stapler, there are multiple occurrences, so it's "fine". Here you do not have a real interest to use `break` instead of just `return` except in cases where you have multiple ErrorCustomizer and wants to use the last one instead of the first one to returns a non-null value.

core/src/main/java/org/kohsuke/stapler/HttpResponses.java
jeffret-b: I'm ambivalent about it, so I'll just leave it as is.

core/src/main/java/org/kohsuke/stapler/HttpResponses.java
jeffret-b: This method doesn't have request, which is used in `RequirePOST` to obtain the class loader. The `HttpResponseException.generateResponse()` method does have the request but that doesn't help us for obtaining the handler.

jeffret-b: There is a coding style that says a method should only ever return from one place. Do you require a different style?

jeffret-b: I don't understand what you want here about "first to handle" and similar patterns. Do you want this to throw an exception, as in `RequirePOST`, instead of returning one?

Wadeck: The request could be retrieved using static call to Stapler, but if it's not required, no need to change.

I am more concerned by the "first to handle" -> in RequirePOST the first error customizer will provode the result while in your proposal, it's the last one.

Of course it works but will cause more headaches if something different happen depending on which customizer you are using.

Let's move forward.

Wadeck: After quick live discussion, I misinterpret the break as a continue and my bad, the behavior was just as the one from Daniel. So good to go.

jeffret-b:
It looks like this is ready to go. @jglick or @daniel-beck, who can help get this merged in and released?

daniel-beck:
> who can help get this merged in and released?

Based on permission files, nobody is currently a maintainer of Stapler, so this can currently only be released by people with universal upload permissions, basically Oliver Gonda and people with Artifactory admin access (Tyler, KK, Wadeck, me). Alternatively, we use part of the security process to deploy into a secondary repo, and then move artifacts over. 😆

(@jglick no longer has his previous universal upload access.)

daniel-beck:
Looks like Stapler is (now?) released to repo.j.o, and nobody has permissions to do this, which basically means Oliver Gondza and the Artifactory admins (KK, Tyler, Wadeck, and me) can do it. Or we use some of the security process to deploy into a private repo and them copy it over. 😆

jglick:
IIUC this is waiting for https://github.com/jenkins-infra/repository-permissions-updater/pull/1420

daniel-beck:
@jglick Looks like the parent POM results in deployment to OSSRH after all? If so, could you release Stapler?

jglick:
https://github.com/stapler/stapler/releases/tag/stapler-parent-1.259

jeffret-b:
Thanks for releasing that, @jglick!

PR #178, opened 2019-12-18
 add new writeLogTo method

liucc52:
add end parameter to avoid large log file cause memory leak


jglick:
Is there an accompanying patch to `jenkinsci/jenkins` which would use this?

jglick:
Not useful without a draft PR in `jenkinsci/jenkins` using the new API.

Issue #181, opened 2020-03-11
IllegalArgumentException on when the setter parameter is of type Object


nfalco79:
The full description of the issue is [here](https://issues.jenkins-ci.org/browse/JENKINS-61438)

In short: RequestImpl instantiate a ChoiceParameterDefinition class and than using reflection lookup a setter `@DataBoundSetter` annotated for "choice" property. It takes the parameter type from method (in this unlikely case Object) and than use apache beanutils to convert the actual value "v1\n2\n3" (String) to Object. Since there are no converter registered for Object it fails

jglick:
[JENKINS-61438](https://issues.jenkins.io/browse/JENKINS-61438)

PR #182, opened 2020-03-11
[JENKINS-61438] IllegalArgumentException for ChoiceParameter on bindJSON saving job configuration

nfalco79:
bindJSON of RequestImpl now returns value itself if no converter found and the expected type is Object.


nfalco79:
This fix #181

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -734,14 +734,23 @@ public Object convertJSON(Object o) {
 
             if (l==null) {// single value conversion
                 Converter converter = Stapler.lookupConverter(type);
-                if (converter==null)
+                if (converter==null) {
```
jglick:
```suggestion
                if (converter == null) {
```

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -734,14 +734,23 @@ public Object convertJSON(Object o) {
...
+                    }
                     throw new IllegalArgumentException("Unable to convert to "+type);
+                }
 
                 return converter.convert(type,o);
             } else {// single value in a collection
                 Converter converter = Stapler.lookupConverter(l.itemType);
-                if (converter==null)
-                    throw new IllegalArgumentException("Unable to convert to "+l.itemType);
+                if (converter==null) {
```
jglick:
```suggestion
                if (converter == null) {
```

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
nfalco79: I will change, I simply keep the actual formatting

jglick: Just old nonstandard code style from @kohsuke; I generally clean up when modifying lines anyway (no extra diff penalty). Does not matter.

nfalco79:
The branch-api-plugin [PR-185](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fbranch-api-plugin/detail/PR-185/1/tests) do not pass the test.
The second commit to use incremental build pass (at least on My PC)

nfalco79:
[x] Added test a case as request
[x] additional test cases to prove use of stapler in a plugin PR
[x] cosmetic changes in PR review

From my side it's ok

nfalco79:
Any chance to get this fixed in next LTS?

jglick:
TBD.

nfalco79:
To Be Do? Done
To Be Decide?

jglick:
Sorry, to be determined, based on when I can make time to do a review and cut a release.

jglick:
Sorry, have not found any time, will try to find another maintainer…

nfalco79:
Ok, please let me know so I can remove the patch from our production

nfalco79:
ping

core/src/test/java/org/kohsuke/stapler/RequestImplTest.java
```diff
@@ -42,15 +47,55 @@
...
+        stapler.setWebApp(new WebApp(Mockito.mock(ServletContext.class)));
+        final MockRequest mockRequest = mockRequest(generateMultipartData());
+        RequestImpl req = new RequestImpl(stapler, mockRequest, Collections.<AncestorImpl>emptyList(), null);
+
+        JSONObject json = new JSONObject();
+        json.put("stapler-class", SetterObject.class.getName());
+        json.put("$class", SetterObject.class.getName());
+        json.put("choices", "1\n2\n3");
+
+        SetterObject o = req.bindJSON(SetterObject.class, json);
```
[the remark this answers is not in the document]
nfalco79: I do not understand what you mean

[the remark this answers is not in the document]
jglick: Sorry, just asking whether this was the line of the test which throws an error if you do not have the code fix in place.

core/src/test/java/org/kohsuke/stapler/RequestImplTest.java
```diff
@@ -42,15 +47,55 @@
...
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ * @author Nikolas Falco
  */
 public class RequestImplTest {
 
+    @Test
+    public void verify_JSON_bind_work_with_setter_that_accept_object_type() throws Exception {
+        final Stapler stapler = new Stapler();
+        stapler.setWebApp(new WebApp(Mockito.mock(ServletContext.class)));
+        final MockRequest mockRequest = mockRequest(generateMultipartData());
```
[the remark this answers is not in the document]
nfalco79: yes it's true, maybe a copy paste from other test. I will remove it

core/src/test/java/org/kohsuke/stapler/RequestImplTest.java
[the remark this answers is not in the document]
nfalco79: yes this line will trigger the `TypePair.convertJSON` where there is the fix. The will be triggered two times:
the first one to build the `SetterObject` class and the second one to build the arguments for the `SetterObject.setChoices` method. Without the changes it will throw a `throw new IllegalArgumentException("Unable to convert to "+type);`

nfalco79:
I did changes as per review and rebase against master

jglick:
> rebase against master

BTW this breaks the ability for reviewers to only look at incremental changes. For a small patch like this, it is not a big deal; for a larger patch it would mean starting the review from scratch.

core/src/test/java/org/kohsuke/stapler/RequestImplTest.java
```diff
@@ -28,29 +29,106 @@
...
+        
+        // Check getParameterNames
+        Assert.assertTrue(Collections.list(request.getParameterNames()).contains("p1"));
+        Assert.assertTrue(Collections.list(request.getParameterNames()).contains("text1"));
+        
+        // Check getParameterMap
+        Assert.assertTrue(request.getParameterMap().containsKey("text1"));        
+    }
+
+    private MockRequest mockRequest(final byte[] buf) {
```
jglick, line 129: So all these changes below could now be reverted?

PR #183, opened 2020-03-13
Correctly encode control and surrogate characters in JSON

daniel-beck:


jglick:
Let me know if you need a release.

daniel-beck:
@jglick Would be nice whenever you have the time to kick it off. Not urgent, so if you want to get https://github.com/stapler/stapler/pull/182 or others over the line first, that's fine. Thanks!

PR #185, opened 2020-05-11
Add spotbugs and findsecbugs

jeffret-b:
Add spotbugs and findsecbugs to Stapler as part of the long, ongoing effort to improve static code analysis. Stapler didn't even have standard spotbugs configured so this PR contains findings relating to general items and to security.

I took it to the Medium level (threshold), which seems to be about right for Stapler. This corrects or points out areas that might be of concern but isn't getting too esoteric.

I suppose there could be an argument that it isn't worthwhile to make these improvements in Stapler, because it isn't in active development. But, we sure seem to make a lot of changes to Stapler. I find that having spotbugs helps us to avoid introducing yet more issues, such as further instances of the public, non-final flag. And it brings better awareness to areas that could be an issue or where we need to tread carefully. Therefore, I believe it is better to introduce this analysis to help better support future work.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
+++ b/core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
@@ -25,9 +25,11 @@ package org.kohsuke.stapler;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+
 import java.net.URL;
 import java.util.logging.Level;
 import java.util.logging.Logger;
--- a/core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
+++ b/core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
@@ -25,6 +25,8 @@ package org.kohsuke.stapler;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -157,6 +159,7 @@ public abstract class AbstractTearOff<CLT,S,E extends Exception> extends Caching
     }
 
     private static final Pattern JAR_URL = Pattern.compile("jar:(file:.+)!/.*");
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "Files are read from approved plugins, not from user input.")
     private static File fileOf(URL res) {
         try {
             switch (res.getProtocol()) {
--- a/core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
+++ b/core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
@@ -25,11 +25,9 @@ package org.kohsuke.stapler;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
-
 import java.net.URL;
 import java.util.logging.Level;
 import java.util.logging.Logger;
--- a/core/src/main/java/org/kohsuke/stapler/ForwardToView.java
+++ b/core/src/main/java/org/kohsuke/stapler/ForwardToView.java
@@ -23,6 +23,8 @@
 
 package org.kohsuke.stapler;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import javax.servlet.ServletException;
 import javax.servlet.RequestDispatcher;
 import java.io.IOException;
@@ -37,8 +39,8 @@ import java.util.Map.Entry;
  * @author Kohsuke Kawaguchi
  */
 public class ForwardToView extends RuntimeException implements HttpResponse {
+    @SuppressFBWarnings(value = "SE_BAD_FIELD", justification = "A valid issue, but complicated to fix now. Not causing any observable problems.")
     private final DispatcherFactory factory;
-    private boolean optional;
     private final Map<String,Object> attributes = new HashMap<String, Object>();
 
     private interface DispatcherFactory {
@@ -86,16 +88,16 @@ public class ForwardToView extends RuntimeException implements HttpResponse {
      * Make this forwarding optional. Render nothing if a view doesn't exist.
      */
     public ForwardToView optional() {
-        optional = true;
         return this;
     }
 
+    @SuppressFBWarnings(value = "REQUESTDISPATCHER_FILE_DISCLOSURE", justification = "Forwarded to a view to handle correctly.")
     public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
         for (Entry<String, Object> e : attributes.entrySet())
             req.setAttribute(e.getKey(),e.getValue());
         RequestDispatcher rd = factory.get(req);
-        if (rd==null && optional)
-            return;
-        rd.forward(req, rsp);
+        if (rd != null) {
+            rd.forward(req, rsp);
+        }
     }
 }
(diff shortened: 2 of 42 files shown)


jeffret-b:
@jvz @daniel-beck @Wadeck

core/src/main/java/org/kohsuke/stapler/TearOffSupport.java
```diff
@@ -37,12 +38,10 @@
  * @author Kohsuke Kawaguchi
  */
 public abstract class TearOffSupport {
-    private volatile Map<Class,Object> tearOffs;
+    private volatile Map<Class,Object> tearOffs = new ConcurrentHashMap<>();
```
jvz, line 41: Chances are that this was lazy initialized to avoid allocating hash map buckets.

jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
```diff
@@ -169,6 +170,11 @@ protected Properties get(String key) {
         return props;
     }
 
+    @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "Not relevant in this situation.")
+    private InputStream openStream(String url) throws IOException {
```
jvz, line 174:
```suggestion
    private static InputStream openStream(String url) throws IOException {
```

jelly/src/main/java/org/kohsuke/stapler/framework/adjunct/AdjunctManager.java
```diff
@@ -163,7 +163,7 @@ public void doDynamic(StaplerRequest req, StaplerResponse rsp) throws IOExceptio
                 return;
             }
             // remember URLs that we can serve. but don't remember error ones, as it might be unbounded
-            allowedResources.put(path,path);
+            allowedResources.putIfAbsent(path,path);
```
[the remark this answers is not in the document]
jeffret-b: AT_OPERATION_SEQUENCE_ON_CONCURRENT_ABSTRACTION : "This code contains a sequence of calls to a concurrent abstraction (such as a concurrent hash map). These calls will not be executed atomically."

The operations of `!allowedResources.containsKey(path)` and `allowedResources.put(path,path)` are not atomic when combined together so there is a potential multithreading issue in the middle of the sequence. The `putIfAbsent` is designed for that purpose.

But, a set is the correct semantic and we can change to `ConcurrentSkipListSet`.

jeffret-b:
Not really sure what to do about these two new findings that were introduced in the recently merged PR introducing a new cache. @jglick any ideas?

core/src/main/java/org/kohsuke/stapler/ForwardToView.java
```diff
@@ -86,16 +88,16 @@ public ForwardToView with(Map<String,?> attributes) {
      * Make this forwarding optional. Render nothing if a view doesn't exist.
      */
     public ForwardToView optional() {
-        optional = true;
```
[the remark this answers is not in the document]
jglick: OK if this was deliberate, just saw a deletion with no explanation in the **Files** tab.

core/src/main/java/org/kohsuke/stapler/MetaClassLoader.java
```diff
@@ -62,6 +64,7 @@ public static MetaClassLoader get(ClassLoader cl) {
     /**
      * If non-null, delegate to this classloader.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_REFACTORED_TO_BE_FINAL", justification = "Legacy switch.")
```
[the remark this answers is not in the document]
jglick: OK, I thought it was non`final` merely because the author had not bothered to factor the initialization expression into a method, which would then be illegal acc. to JLS even though it is not mutated after class initialization.

core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -314,6 +316,11 @@ public long getLastModified() {
                 return connection.getLastModified();
             }
         }
+
+        @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "Protected by checks at other layers.")
+        private long determineLastModified(URL jarURL) {
+            return new File(jarURL.getFile()).lastModified();
```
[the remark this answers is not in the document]
jglick: Please do so.

jelly/src/main/java/org/kohsuke/stapler/framework/adjunct/AdjunctManager.java
```diff
@@ -157,13 +158,13 @@ public void doDynamic(StaplerRequest req, StaplerResponse rsp) throws IOExceptio
...
 
-        if(!allowedResources.containsKey(path)) {
+        if(!allowedResources.contains(path)) {
             if(!allowResourceToBeServed(path)) {
                 rsp.sendError(SC_FORBIDDEN);
                 return;
             }
             // remember URLs that we can serve. but don't remember error ones, as it might be unbounded
-            allowedResources.put(path,path);
+            allowedResources.add(path);
```
jglick, line 167: No this does not fix the issue at all—it is still doing _two_ operations. That is why I suggested using the return value of `add`: it atomically adds the path if not already present, and lets you know if it did so or not.

jelly/src/main/java/org/kohsuke/stapler/framework/adjunct/AdjunctManager.java
```diff
@@ -157,13 +158,13 @@ public void doDynamic(StaplerRequest req, StaplerResponse rsp) throws IOExceptio
...
 
-        if(!allowedResources.containsKey(path)) {
+        if(!allowedResources.contains(path)) {
             if(!allowResourceToBeServed(path)) {
                 rsp.sendError(SC_FORBIDDEN);
                 return;
             }
             // remember URLs that we can serve. but don't remember error ones, as it might be unbounded
-            allowedResources.putIfAbsent(path,path);
+            allowedResources.add(path);
```
jglick: as discussed, use return value from `add` in conditional

core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -645,6 +659,7 @@ private String getMimeType(String fileName) {
      *
      * See http://weblogs.java.net/blog/kohsuke/archive/2007/04/how_to_convert.html
      */
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "Protected by checks at other layers.")
     /*package for test*/ File toFile(URL url) {
         String urlstr = url.toExternalForm();
         if(!urlstr.startsWith("file:"))
```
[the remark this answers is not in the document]
jglick: at least leave a `TODO`

jglick:
> two new findings that were introduced in the recently merged PR introducing a new cache

Oh I see, new `PATH_TRAVERSAL_IN`s to suppress (reading files from plugin code not users).

core/src/main/java/org/kohsuke/stapler/AbstractTearOff.java
```diff
@@ -25,9 +25,11 @@
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+
```
jglick, line 32: Why?

core/src/main/java/org/kohsuke/stapler/HttpRedirect.java
```diff
@@ -23,6 +23,8 @@
 
 package org.kohsuke.stapler;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
```
jglick, line 27:
```suggestion
```

jeffret-b:
Builds fully again with another new suppression.

Wadeck:
It seems the PR is already reviewed by two persons, so I do not plan to add another review.

jeffret-b:
Hmm ... it's passing locally. Let's see what a rebuild does.

jeffret-b:
Build is passing now.

jelly/src/main/java/org/kohsuke/stapler/framework/adjunct/AdjunctManager.java
jglick: OK, needed to reread

> don't remember error ones, as it might be unbounded

IOW this is not a regular cache and cannot use the regular idioms.

PR #197, opened 2020-12-13
Don't use setAccessible, fixes illegal reflective access warnings

timja:
Some simple testing seems to show this is ok

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
@@ -6,6 +6,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -87,7 +88,7 @@ public abstract class FieldRef extends AnnotatedRef {
                 try {
                     return f.get(instance);
                 } catch (IllegalAccessException e) {
-                    LOGGER.log(Level.WARNING, "please report to the respective component", e);
+                    LOGGER.log(Level.WARNING, "Please report to the respective component", e);
                     f.setAccessible(true);
                     return f.get(instance);
                 }
--- a/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
@@ -82,7 +82,12 @@ public abstract class FieldRef extends AnnotatedRef {
 
             @Override
             public Object get(Object instance) throws IllegalAccessException {
-                return f.get(instance);
+                try {
+                    return f.get(instance);
+                } catch (IllegalAccessException e) {
+                    f.setAccessible(true);
+                    return f.get(instance);
+                }
             }
 
             @Override
--- a/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
@@ -69,8 +69,6 @@ public abstract class FieldRef extends AnnotatedRef {
     }
 
     public static FieldRef wrap(final Field f) {
-        f.setAccessible(true);
-
         return new FieldRef() {
             @Override
             public <T extends Annotation> T getAnnotation(Class<T> type) {
--- a/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
@@ -6,6 +6,8 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 /**
@@ -15,6 +17,8 @@ import java.util.stream.Collectors;
  * @author Kohsuke Kawaguchi
  */
 public abstract class FieldRef extends AnnotatedRef {
+
+    private static final Logger LOGGER = Logger.getLogger(FieldRef.class.getName());
 
     public interface Filter {
         boolean keep(FieldRef m);
@@ -85,6 +89,7 @@ public abstract class FieldRef extends AnnotatedRef {
                 try {
                     return f.get(instance);
                 } catch (IllegalAccessException e) {
+                    LOGGER.warning(e.getMessage() + ", please report to the respective component");
                     f.setAccessible(true);
                     return f.get(instance);
                 }
--- a/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
@@ -87,7 +87,7 @@ public abstract class FieldRef extends AnnotatedRef {
                 try {
                     return f.get(instance);
                 } catch (IllegalAccessException e) {
-                    LOGGER.warning(e.getMessage() + ", please report to the respective component");
+                    LOGGER.log(Level.WARNING, "please report to the respective component", e);
                     f.setAccessible(true);
                     return f.get(instance);
                 }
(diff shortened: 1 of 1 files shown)


daniel-beck:
As a first change, could we determine when this call would make a difference and report instances e.g. through Uplink telemetry?

timja:
> As a first change, could we determine when this call would make a difference and report instances e.g. through Uplink telemetry?

alternatively we could do try catch set accessible and try again ?

daniel-beck:
IIUC, this is a deprecation warnings for a feature that's going away in future Java releases. So we need to be able to tell when we rely on this behavior. A silent fallback, while compatible, doesn't do that. I recommend that we log a warning or set up a hook into telemetry so that we or users can file issues with affected components.

timja:
> IIUC, this is a deprecation warnings for a feature that's going away in future Java releases. So we need to be able to tell when we rely on this behavior. A silent fallback, while compatible, doesn't do that. I recommend that we log a warning or set up a hook into telemetry so that we or users can file issues with affected components.

You got a link with telemetry example?

daniel-beck:
> You got a link with telemetry example?

Core has some examples in https://github.com/jenkinsci/jenkins/tree/master/core/src/main/java/jenkins/telemetry/impl

It's annoying from Stapler, I basically defined https://github.com/stapler/stapler/blob/96af01cc7f61b7223946f1580d59eb72991a3eed/core/src/main/java/org/kohsuke/stapler/EvaluationTrace.java#L71-L103 and then implemented that in core at https://github.com/jenkinsci/jenkins/blob/f7884d3513c6e436baef432a486197a3bb4d846c/core/src/main/java/jenkins/telemetry/impl/StaplerDispatches.java#L101-L112

Logging would be the lower effort approach.

core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
```diff
@@ -84,7 +84,13 @@ public String getName() {
 
             @Override
             public Object get(Object instance) throws IllegalAccessException {
-                return f.get(instance);
+                try {
+                    return f.get(instance);
+                } catch (IllegalAccessException e) {
+                    LOGGER.warning(e.getMessage() + ", please report to the respective component");
```
jglick, line 90:
```suggestion
                    LOGGER.log(Level.WARNING, "please report to the respective component", e);
```

jglick (approved):
Sounds OK. I do not think this should have been used on private fields to begin with?

jglick:
> Logging would be the lower effort approach

Indeed, telemetry feels like overkill here.

jglick:
Just let me know if/when you want a release—I am able to cut one. (Not trivial.)

timja:
> Just let me know if/when you want a release—I am able to cut one. (Not trivial.)

I would like a release please 😄 .

When you have time

core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
```diff
@@ -84,7 +85,13 @@ public String getName() {
 
             @Override
             public Object get(Object instance) throws IllegalAccessException {
-                return f.get(instance);
+                try {
+                    return f.get(instance);
+                } catch (IllegalAccessException e) {
+                    LOGGER.log(Level.WARNING, "Please report to the respective component", e);
```
daniel-beck, line 91: Does this log enough details to be useful to users?

timja:
```
Dec 14, 2020 3:45:23 PM org.kohsuke.stapler.lang.FieldRef$1 get
WARNING: please report to the respective component
java.lang.IllegalAccessException: class org.kohsuke.stapler.lang.FieldRef$1 cannot access a member of class org.kohsuke.stapler.AncestorImplTest$Foo with modifiers "public"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:361)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:591)
	at java.base/java.lang.reflect.Field.checkAccess(Field.java:1075)
	at java.base/java.lang.reflect.Field.get(Field.java:416)
	at org.kohsuke.stapler.lang.FieldRef$1.get(FieldRef.java:89)
	at org.kohsuke.stapler.MetaClass$1.doDispatch(MetaClass.java:170)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:58)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:766)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
	at org.kohsuke.stapler.MetaClass$1.doDispatch(MetaClass.java:170)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:58)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:766)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:694)
	at org.kohsuke.stapler.Stapler.service(Stapler.java:240)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:487)
	at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java:362)
	at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java:181)
	at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java:729)
	at org.mortbay.jetty.handler.HandlerWrapper.handle(HandlerWrapper.java:152)
	at org.mortbay.jetty.Server.handle(Server.java:324)
	at org.mortbay.jetty.HttpConnection.handleRequest(HttpConnection.java:505)
	at org.mortbay.jetty.HttpConnection$RequestHandler.headerComplete(HttpConnection.java:829)
	at org.mortbay.jetty.HttpParser.parseNext(HttpParser.java:513)
	at org.mortbay.jetty.HttpParser.parseAvailable(HttpParser.java:211)
	at org.mortbay.jetty.HttpConnection.handle(HttpConnection.java:380)
	at org.mortbay.jetty.bio.SocketConnector$Connection.run(SocketConnector.java:228)
	at org.mortbay.thread.QueuedThreadPool$PoolThread.run(QueuedThreadPool.java:488)

```

timja: hmm I was intentionally not logging the stacktrace before as it was useless, but @jglick's suggestion added it, shall i revert to just a log line?

jglick:
I tried to cut a release but I am afraid I was unable to do so; oss.sonatype.org is timing out trying to close the staging repo. Under the best of circumstances releasing this repo is a very slow and flaky process.

Since Stapler is AFAIK only really used inside Jenkins, it would be preferable to deploy only to our Artifactory (maybe also move it inside @jenkinsci), but not sure who has the permissions to make that happen.

timja:
> Since Stapler is AFAIK only really used inside Jenkins, it would be preferable to deploy only to our Artifactory (maybe also move it inside @jenkinsci), but not sure who has the permissions to make that happen.

anyone can add the files required for artifactory to take it, just add the file here: https://github.com/jenkins-infra/repository-permissions-updater/tree/master/permissions

and update the pom configuration

jglick:
It is already there

```yaml
# regularly published to OSSRH, this file exists only for incrementals support
```

so this would be a task to change `pom.xml` config to publish to Artifactory (a bunch would be automatic by switching to `jenkinsci/bom`) _and_ to look around the Internet for people still using Stapler in other contexts and see if they can be notified that it is going away or at least no further updates will appear on Central. The former I can probably do pretty easily. Volunteers for the latter?

There is also the `github-pages` MRP publisher which is very slow and awful which it would be good to avoid but http://stapler.kohsuke.org/ still uses this and would need to become sort of redirect to static Markdown/AsciiDoc pages I guess, plus maybe something in javadoc.jenkins.io and maybe just remove the generated tag reference.

daniel-beck:
While the release isn't out, could we hold until we've amended the log message to be more useful?

core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
jglick: > Does this log enough details

Includes class name of the field, so seems OK to me.

> shall i revert to just a log line?

You can, no strong opinion.

timja:
> While the release isn't out, could we hold until we've amended the log message to be more useful?

https://github.com/stapler/stapler/pull/199

PR #199, opened 2020-12-14
Quieter log

timja:
```
WARNING: java.lang.IllegalAccessException: Please report to the respective component:
 class org.kohsuke.stapler.lang.FieldRef$1 cannot access a member of class org.kohsuke.stapler.AncestorImplTest$Foo with modifiers "public"
```

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
+++ b/core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
@@ -6,7 +6,6 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -88,7 +87,7 @@ public abstract class FieldRef extends AnnotatedRef {
                 try {
                     return f.get(instance);
                 } catch (IllegalAccessException e) {
-                    LOGGER.log(Level.WARNING, "Please report to the respective component", e);
+                    LOGGER.warning(e.getClass().getName() + ": Please report to the respective component:\n " + e.getMessage());
                     f.setAccessible(true);
                     return f.get(instance);
                 }


jglick (approved):

core/src/main/java/org/kohsuke/stapler/lang/FieldRef.java
```diff
@@ -88,7 +87,7 @@ public Object get(Object instance) throws IllegalAccessException {
                 try {
                     return f.get(instance);
                 } catch (IllegalAccessException e) {
-                    LOGGER.log(Level.WARNING, "Please report to the respective component", e);
+                    LOGGER.warning(e.getClass().getName() + ": Please report to the respective component:\n " + e.getMessage());
```
daniel-beck, line 90: @timja I'm confused, this message isn't any useful to admins encountering it than the previous one? I was thinking it should explain that nothing breaks on Java 8 and 11 but would in newer Java releases, that the class name indicates what the problem is, etc.

In the past, in similar situations, we've often used a `jenkins.io/redirect` URL to link to further documentation and allow updating after release. Obviously, this is Stapler, not Jenkins, so perhaps it could be more generic than that?

jglick: > perhaps it could be more generic

:-1: may as well assume this is for Jenkins

jglick: I am not sure how much effort we want to put into enhancing the error message. Really we want Jenkins core/plugin _developers_ to see this during testing and deal with it. If some intrepid admin happens across such a warning and wants to get into a fix, great, but this is not expected.

daniel-beck: 🤷 Works for me, but then we should be a lot more specific than we are, e.g. why call it component when it's almost certainly a plugin?

timja: It could be some part of core too, I didn’t find any when doing some experimenting. I didn’t want to assume it was a plugin... and yeah agree with Jesse.

If you have better wording 10000% we can add it

daniel-beck: > If some intrepid admin happens across such a warning and wants to get into a fix, great, but this is not expected.

Isn't the default for any existing code going to be that people will update Jenkins weeklies and encounter this? Why would it be the developers?

Regarding phrasing, how about:


```suggestion
                    LOGGER.warning(e.getClass().getName() + ": Processing this request relies on deprecated behavior that will be disallowed in future releases of Java. See https://jenkins.io/redirect/stapler-reflective-access/ for more information. Details: " + e.getMessage());
```

Then a fairly simple page under https://www.jenkins.io/doc/developer/handling-requests/ can explain the issue and fix.

timja: https://github.com/stapler/stapler/pull/200

PR #201, opened 2020-12-16
Handle more cases of illegal reflective access

timja:
MethodRef could be triggered on the Jenkins core configuration page before

After moving the setAccessible to not the default I haven't been able to trigger it

Databoundsetter i haven't been able to trigger an error with but figured it made sense to make that change

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -40,6 +40,7 @@ import org.jvnet.tiger_types.Lister;
 import org.kohsuke.stapler.bind.BoundObjectTable;
 import org.kohsuke.stapler.lang.Klass;
 import org.kohsuke.stapler.lang.MethodRef;
+import org.kohsuke.stapler.util.IllegalReflectiveAccessLogHandler;
 
 import javax.servlet.RequestDispatcher;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -824,8 +825,13 @@ private Object bindResolve(Object o, JSONObject src) {
...
                             if (f.getAnnotation(DataBoundSetter.class)!=null) {
-                                f.setAccessible(true);
-                                f.set(r, bindJSON(f.getGenericType(), f.getType(), j.get(key)));
+                                try {
+                                    f.set(r, bindJSON(f.getGenericType(), f.getType(), j.get(key)));
+                                } catch(IllegalAccessException e) {
+                                    LOGGER.warning(IllegalReflectiveAccessLogHandler.get(e));
+                                    f.setAccessible(true);
+                                    f.set(r, bindJSON(f.getGenericType(), f.getType(), j.get(key)));
+                                }
```
jglick, line 834: Could move this idiom into a helper method in `IllegalReflectiveAccessLogHandler`.

jglick:
FYI I finally managed to release this stuff: https://github.com/stapler/stapler/releases/tag/stapler-parent-1.262

PR #203, opened 2021-01-09
Use Java 7 versions of `GZIPInputStream` and `GZIPOutputStream`

basil:
See jenkinsci/workflow-api-plugin#134, jenkinsci/cloudbees-folder-plugin#174, and jenkinsci/docker-build-step-plugin#75. [JZlib](https://github.com/ymnk/jzlib) has not been updated since 2013. Since then, [`GZIPInputStream`](https://docs.oracle.com/javase/8/docs/api/java/util/zip/GZIPInputStream.html) and [`GZIPOutputStream`](https://docs.oracle.com/javase/8/docs/api/java/util/zip/GZIPOutputStream.html) have been added to the Java 7 API, so it seems more prudent to use those implementations.

To verify the correctness of the change to `Stapler#isSocketException(Throwable)`, compare the upstream error messages from [JZlib](https://github.com/ymnk/jzlib/blob/a21be20213d66eff15904d925e9b721956a01ef7/src/main/java/com/jcraft/jzlib/DeflaterOutputStream.java#L90) and [AdoptOpenJDK](https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/9a91972c76ddda5c1ce28b50ca38cbd8a30b7a72/jdk/src/share/classes/java/util/zip/DeflaterOutputStream.java#L201).

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -16,11 +16,6 @@
   </properties>
 
   <dependencies>
-    <dependency>
-      <groupId>com.jcraft</groupId>
-      <artifactId>jzlib</artifactId>
-      <version>1.1.3-kohsuke-1</version>
-    </dependency>
     <!-- optional REST support -->
     <dependency>
       <groupId>com.sun.xml.txw2</groupId>
--- a/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/ResponseImpl.java
@@ -23,7 +23,6 @@
 
 package org.kohsuke.stapler;
 
-import com.jcraft.jzlib.GZIPOutputStream;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.ByteArrayOutputStream;
@@ -40,6 +39,7 @@ import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.zip.GZIPOutputStream;
 import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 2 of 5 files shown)


basil:
Oh wow, we actually switched from the Java API _to_ JZlib in [JENKINS-14362](https://issues.jenkins.io/browse/JENKINS-14362) for performance reasons. Well never mind, then!

jvz:
That might be worth benchmarking again. The JIT compiler in the JRE has added so many optimization features over the past several years that some algorithms run faster in pure Java compared to linking to native code due to things like link-time optimization and other techniques that are only _just_ starting to make their way into C and C++ compilers in limited support. (i.e., `javac` has more information to work with compared to something like `ld` which has to link relocatable binaries which can only optimize to their specified boundaries rather than the entire application amongst other JIT techniques for spending time optimizing hot paths).

basil:
Interestingly enough, the Java API is a wrapper around the C-based `zlib(3)`, while _JZlib_ is a pure Java implementation. So any JIT optimizations would probably benefit JZlib more. Reading JENKINS-14362 more closely, it appears that there wasn't a performance problem per se but rather a problem with threads getting stuck in native `zlib(3)` code forever, eventually consuming all CPU cores. I doubt it would be possible to verify that we aren't regressing the fix for JENKINS-14362 since nobody ever figured out how to reproduce the symptoms of JENKINS-14362. The problem in the Java standard library and/or `zlib(3)` described in JENKINS-14362 may have been resolved in the intervening years, but it may very well still be present. I checked the commit logs for the relevant Java and `zlib(3)` code but couldn't come up with any convincing proof that this bug has been fixed. [JDK-8193682](https://bugs.openjdk.java.net/browse/JDK-8193682) (from the 2017/2018 timeframe) describes a very similar issue and doesn't give me confidence. I think the only way to know for sure would be to release the change and see if any users complain about a regression. IMHO the benefit does not seem to be worth the potential inconvenience to users.

MarkEWaite (approved):

PR #205, opened 2021-01-21
@AnyVerb annotation makes web methods explicitly accept any HTTP verb

daniel-beck:
I want an explicit version of the implied default. Alternatively, people could use `@GET @POST` on the same method (more verbs we rarely care about).

WIP. Untested.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/verb/AnyVerb.java
+++ b/core/src/main/java/org/kohsuke/stapler/verb/AnyVerb.java
@@ -1,0 +1,26 @@
+package org.kohsuke.stapler.verb;
+
+import org.kohsuke.stapler.WebMethod;
+import org.kohsuke.stapler.interceptor.InterceptorAnnotation;
+import org.kohsuke.stapler.interceptor.Stage;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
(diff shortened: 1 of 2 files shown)


jglick (commented):
Why would we do this? Seems to be just encouraging people to do the wrong thing.

daniel-beck:
> Why would we do this? Seems to be just encouraging people to do the wrong thing.

In cases where endpoints integrate with code that sends GET requests (IIRC, all the `doFill`s, or similar – something like that blocked my progress on https://github.com/jenkinsci/jenkins/pull/4623), or for endpoints that distinguish between GET and POST inside the method. Tons of code like that in core, e.g. all the `doConfigDotXml`. Can that be done more nicely since the verb annotations were introduced? Sure. This still has value in a transition phase.

jglick:
We should do https://github.com/jenkinsci/jenkins/pull/4623 and then there should be few endpoints I can think of that actually accept both.
* `doConfigDotXml` as you mentioned; rarely defined in plugin code (normally inherited from `AbstractItem`)
* `doBuild` taking an optional token; only redefined in `build-token-root-plugin`

jglick:
Oh and `doCommitStatus` or similar in some SCM plugins historically allowing anonymous `GET` requests to be used from post-commit hooks. (Requiring `POST` and add a `CrumbExclusion` is preferred going forward.)

PR #206, opened 2021-01-31
Removing usages of Guava from tests

basil:


jglick:
Even after this, #207, and #208, it seems there are still some usages of `com.google.common.collect`, `com.google.common.cache`, and `com.google.common.util.concurrent`. Do you plan to clean up those as well?

basil:
Not in the short term. In the medium term, I might look into it. Perhaps [Caffeine](https://github.com/ben-manes/caffeine) would be a good replacement for Guava's cache?

jglick:
Or just `HashMap.computeIfAbsent` etc., if the needs are not exotic.

PR #209, opened 2021-02-02
Bump Commons BeanUtils from 1.9.2 to 1.9.4

basil:
Bumps [Commons BeanUtils](https://commons.apache.org/proper/commons-beanutils/) from 1.9.2 to 1.9.4.

See [the release notes](https://commons.apache.org/proper/commons-beanutils/javadocs/v1.9.4/RELEASE-NOTES.txt). Starting in 1.9.4, Commons BeanUtils does not allow class level access by default. However, our existing Jelly templates rely on class level access. Therefore, we opt out of the default behavior in stapler/stapler#209, following the recommendation in the release notes:

>If one would like to opt out of the default behaviour, one could follow the example set out in the test class available in [`src/test/java/org/apache/commons/beanutils/bugs/Jira520TestCase.java`](https://github.com/apache/commons-beanutils/blob/23f313c0fa2ee8caa3021ab53dd3925b64184f7f/src/test/java/org/apache/commons/beanutils2/bugs/Jira520TestCase.java#L49-L50).

I added this code to [`JellyFacet`](https://github.com/basil/stapler/blob/3a93485b49597c44ae4e4e1d5c767a022522a691/jelly/src/main/java/org/kohsuke/stapler/jelly/JellyFacet.java#L69-L75), where it runs as early as possible during Jelly initialization:

```
<init>:67, JellyFacet (org.kohsuke.stapler.jelly)
[…]
discoverExtensions:105, Facet (org.kohsuke.stapler)
<init>:167, WebApp (org.kohsuke.stapler)
get:69, WebApp (org.kohsuke.stapler)
<init>:924, Jenkins (jenkins.model)
<init>:85, Hudson (hudson.model)
[…]
```

PR #211, opened 2021-03-03
Update to commons-beanutils:1.9.4 without disabling the protection

daniel-beck:
This pull request attempts to update to `commons-beanutils:1.9.4` without disabling the additional protection it offers around the `class` attribute.

It seems in my (fairly limited so far) testing that the problem is `st:include`'s `class` attribute. So this pull request renames the attribute to `clazz` (and keeps the existing setter for compatibility).

This PR additionally proposes a workaround that applies when Jelly files are being parsed and the resulting tags are created: For the `st:include` tag, when the `class` attribute would be set, instead set `clazz`. This attempts to take care of all existing code that sets the `class` attribute for this tag.

I am unsure how best to go about ensuring this doesn't cause a lot of problems. While this fix looks fairly safe, it may also be incomplete. In the short term, #209 seems like the safer approach.

CC @basil

### DIFF
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
@@ -18,6 +18,9 @@ import org.apache.commons.jelly.impl.TagScript;
 
     @Override
     public void addAttribute(String name, Expression expression) {
+        if (replacement.equals(name)) {
+            throw new IllegalArgumentException("The attribute '" + replacement + "' cannot be set on this element, set '" + original + "' instead");
+        }
         if (original.equals(name)) {
             super.addAttribute(replacement, expression);
         } else {
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
@@ -19,7 +19,8 @@ import org.apache.commons.jelly.impl.TagScript;
     @Override
     public void addAttribute(String name, Expression expression) {
         if (replacement.equals(name)) {
-            throw new IllegalArgumentException("The attribute '" + replacement + "' cannot be set on this element, set '" + original + "' instead");
+            // cf. TagScript#run
+            throw new IllegalArgumentException("This tag does not understand the '" + replacement + "' attribute");
         }
         if (original.equals(name)) {
             super.addAttribute(replacement, expression);
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
@@ -1,0 +1,27 @@
+package org.kohsuke.stapler.jelly;
+
+import org.apache.commons.jelly.expression.Expression;
+import org.apache.commons.jelly.impl.TagScript;
+
+/**
+ * This class implements a {@link TagScript} that allows rewriting attribute names.
+ *
+ */
+/* package */ class AttributeNameRewritingTagScript extends TagScript {
+    private String original;
+    private final String replacement;
+
+    public AttributeNameRewritingTagScript(String original, String replacement) {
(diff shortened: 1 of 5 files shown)


jelly/src/main/java/org/kohsuke/stapler/jelly/AttributeNameRewritingTagScript.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.apache.commons.jelly.expression.Expression;
+import org.apache.commons.jelly.impl.TagScript;
+
+/**
+ * This class implements a {@link TagScript} that allows rewriting attribute names.
+ *
+ */
+/* package */ class AttributeNameRewritingTagScript extends TagScript {
+    private String original;
```
basil, line 11:
```suggestion
    private final String original;
```

basil (approved):
Looks great!

jelly/src/main/java/org/kohsuke/stapler/jelly/IncludeTag.java
```diff
@@ -79,8 +84,18 @@ public void setFrom(Object from) {
...
+     * problems with new commons-beanutils restrictions via
+     * {@code ConvertingWrapDynaBean} use in {@code JellyBuilder}.
      */
+    public void setClazz(Class clazz) {
+        this.clazz = clazz;
+    }
+
+    @Deprecated // TODO Remove this method?
     public void setClass(Class clazz) {
+        LOGGER.log(Level.WARNING, "Unexpected call to #setClass", new Exception());
```
basil, line 98: FYI this warning might be a bit noisy. IIUC plugin maintainers won't be able to migrate to the new syntax until after they update their baseline to the first version of Jenkins core where this is released, since if they need to support older versions of Jenkins core they need to keep using the old syntax in their Jelly files. Besides, I found 1,070 unique instances of `<st:include class=""/>` attributes that needed updating. I doubt that log files and Jira tickets are going to help speed this migration along. Rather I wrote a quick and dirty tool at https://github.com/basil/jelly-migrator/ that can be used once a plugin's Jenkins core baseline is updated to automatically update all Jelly files to the new syntax.

daniel-beck: See https://github.com/jenkinsci/jenkins/pull/5324#issuecomment-790371595

Because of the migration code in `StaplerTagLibrary` I expect there to be absolutely no callers left.

rsandell (approved):
We've had riskier workarounds in the past.

jelly/src/main/java/org/kohsuke/stapler/jelly/IncludeTag.java
daniel-beck: Added escape hatch that would allow admins to disable this warning.

MRamonLeon (approved):

jeffret-b (approved):
This makes good sense and is the best proposal about how to move forward and upgrade to the new version. We can't just allow ourselves to remain on the old version and be blocked from continuing forward.

Hopefully this will work. If it doesn't there is a sufficient amount of escape hatches and configuration to allow people with other scenarios to continue while we correct anything additional that might arise.

basil (approved):
This looks great! Very nice change.

oleg-nenashev (approved):
Looks right. Kudos to the Jelly team and @jstrachan for such extensibility

alecharp (approved):

MRamonLeon:
Who's able to merge this PR @daniel-beck, @oleg-nenashev? It has enough approvals for a few days already. The permission updater repo doesn't mention any maintainer: https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/component-stapler.yml

daniel-beck:
> The permission updater repo doesn't mention any maintainer

That's explained in the linked file in a comment.

@jglick Please hold off releasing for another week until we've merged in changes from the upcoming security update.

batmat:
Why can't we just release and hold-off updating this into Jenkins Core?
We do not understand the rationale here. Thanks!

daniel-beck:
Because there's a privately staged new release of Stapler. That release you're asking for would never be in a core update anyway (and if it were it'd be a pretty big regression over next week's core release for absolutely no reason).

PR #213, opened 2021-04-15
Remove Stapler's Guava dependency

basil:
### Motivation

This is a step along the way to possibly being able to shade the Guava dependency in Jenkins core as described in [jenkinsci/jenkins#5059 (comment)](https://github.com/jenkinsci/jenkins/pull/5059#issuecomment-763290152). I'm not yet sure that effort will be feasible, but this change fell out of that effort and seems to have independent value on its own. At best, this change might be helpful to the effort to modernize Guava in Jenkins core; at worst, this change slims down Stapler's dependency tree, which isn't bad in and of itself.

### Guava replacements

I dealt with the three remaining Guava dependencies in Stapler as follows:

- For usages of Guava's cache, I replaced them with usages of Caffeine or `Map#computeIfAbsent` if the needs were not exotic as suggested in [stapler/stapler#206 (comment)](https://github.com/stapler/stapler/pull/206#issuecomment-789169206). Mercifully, I didn't have to use Caffeine in the `jelly/` submodule, so I only had to introduce a Caffeine dependency for the `core/` Stapler module.
- For usages of `Iterators#limit` I simply inlined the corresponding Guava code (with the appropriate license).

### Shading Caffeine

The above having been done, I was concerned about exposing Caffeine as a new dependency. I'm not sure how stable its API is, and including it as a transitive dependency of Jenkins core implies that we are committing to supporting that API for all plugins in the future. It seems safer to avoid making that commitment unless it is truly necessary. So I tried to figure out how to shade the Caffeine dependency so that its usage remains internal to Stapler.

Unfortunately, using the Maven Shade Plugin in reactor builds (including both Stapler and Jenkins core) is a bit of a mess due to [MNG-5899](https://issues.apache.org/jira/browse/MNG-5899). In particular, the so-called "dependency reduced POM" doesn't quite work right in reactor builds. After spending quite a bit of time trying to understand how this works (or doesn't) and scratching my head, I decided to dispense with this unreliable "magic" and set `createDependencyReducedPom` to false. While this does have some downsides, the upside is that at least one doesn't have to deal with the pathological behavior of MNG-5899. With `createDependencyReducedPom` set to false, we produce a Stapler JAR with the shaded Caffeine classes included (good), without a drastically rewritten POM file (good in the sense that it avoids the risk of a regression), predictable behavior in reactor builds (good), and a dependency on Caffeine in the POM file (bad in the sense that the classes are already shaded so this is unnecessary). The main downside is the unnecessary dependency on Caffeine in the POM. That can be excluded on the consumer side, which is ugly, but it works.

Overall this is the simplest approach I could come up with that actually works. I tried a lot of other approaches that didn't actually work. I recognize this isn't necessarily the most elegant solution, but I don't think there really are any elegant ways of using the Maven Shade Plugin in reactor builds. If anyone can provide a proof of concept for a different way of shading Caffeine that actually works and is more elegant, I would be happy to go with that instead.

### Testing done

I ran the Jenkins core test suite against this change in jenkinsci/jenkins#5422. Note that there I am excluding the Caffeine dependency, which is declared in the POM file (because `createDependencyReducedPom` is false) but not actually needed (because the Caffeine classes are shaded in the core Stapler JAR file).

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Function.java
+++ b/core/src/main/java/org/kohsuke/stapler/Function.java
@@ -23,9 +23,9 @@
 
 package org.kohsuke.stapler;
 
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.CacheLoader;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.interceptor.Interceptor;
 import org.kohsuke.stapler.interceptor.InterceptorAnnotation;
@@ -195,7 +195,7 @@ public abstract class Function {
                 }
 
                 // if the databinding method is provided, call that
-                Function binder = PARSE_METHODS.getUnchecked(t);
+                Function binder = PARSE_METHODS.get(t);
                 if (binder!=RETURN_NULL) {
                     arguments[i] = binder.bindAndInvoke(null,req,rsp);
                     continue;
@@ -226,7 +226,7 @@ public abstract class Function {
             throw new AssertionError(e);    // impossible
         }
 
-        PARSE_METHODS = CacheBuilder.newBuilder().weakKeys().build(new CacheLoader<Class,Function>() {
+        PARSE_METHODS = Caffeine.newBuilder().weakKeys().build(new CacheLoader<Class,Function>() {
             public Function load(Class from) {
                 // MethdFunction for invoking a static method as a static method
                 FunctionList methods = new ClassDescriptor(from).methods.name("fromStapler");
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
@@ -23,9 +23,6 @@
 
 package org.kohsuke.stapler.jelly;
 
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import org.apache.commons.jelly.JellyContext;
 import org.apache.commons.jelly.JellyException;
 import org.apache.commons.jelly.TagLibrary;
@@ -35,6 +32,8 @@ import org.kohsuke.stapler.MetaClassLoader;
 
 import java.lang.ref.WeakReference;
 import java.net.URL;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * {@link MetaClassLoader} tear-off for Jelly support.
@@ -47,7 +46,7 @@ public class JellyClassLoaderTearOff {
     /**
      * See {@link JellyClassTearOff#scripts} for why we use {@link WeakReference} here.
      */
-    private volatile WeakReference<LoadingCache<String,TagLibrary>> taglibs;
+    private volatile WeakReference<ConcurrentMap<String,TagLibrary>> taglibs;
 
     static ExpressionFactory EXPRESSION_FACTORY = new JexlExpressionFactory();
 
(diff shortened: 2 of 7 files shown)


daniel-beck:
Shading is configured but unused, or what am I missing?

basil:
I wrote a description of the change that hopefully answers your questions. Feel free to let me know if it doesn't or if there's anything else I can clarify.

daniel-beck:
@basil Thanks, now I think I see the relation between your description and the code. The only shading that seemed to really work in the past is adding a new module that just creates a shaded library, then using that, and that'll be reflected in downstream source code, which is what I expected to see here.

core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -226,7 +226,7 @@ Object bindAndInvoke(Object o, StaplerRequest req, StaplerResponse rsp, Object..
             throw new AssertionError(e);    // impossible
         }
 
-        PARSE_METHODS = CacheBuilder.newBuilder().weakKeys().build(new CacheLoader<Class,Function>() {
+        PARSE_METHODS = Caffeine.newBuilder().weakKeys().build(new CacheLoader<Class,Function>() {
```
jglick, line 229: I doubt this ever worked to begin with. See https://github.com/jglick/biweak-class-map

jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
```diff
@@ -56,46 +55,43 @@ public JellyClassLoaderTearOff(MetaClassLoader owner) {
...
+            m = new ConcurrentHashMap<>();
+            taglibs = new WeakReference<>(m);
+        }
+        TagLibrary tl = m.computeIfAbsent(nsUri, key -> {
                     if(owner.parent!=null) {
                         // parent first
-                        TagLibrary tl = owner.parent.loadTearOff(JellyClassLoaderTearOff.class).getTagLibrary(nsUri);
-                        if(tl!=null)    return tl;
+                        TagLibrary taglib = owner.parent.loadTearOff(JellyClassLoaderTearOff.class).getTagLibrary(key);
+                        if(taglib!=null)    return taglib;
```
jglick, line 69: Avoid variable renaming not directly needed, to minimize patch.

jglick (commented):
IIUC Caffeine is really needed here just for a couple of things using `softValues`. I wonder whether we want to keep this logic.

jglick:
@jtnord PTAL

jtnord:
So Caffeine within major lines has good backwards compatibility.  Version 3.x breaks 2.x API compatibility and does so using the same `artifactId` and same package names - so if we do not shade we will have stored an issue in the future (although the breakage is much less than the guava API bump breakage).

I have been thinking about how to deal with Guava and the cache in `config-as-code` plugin (actually more generally in Jenkins and this probably deserves a dump of my thoughts into the Jenkins core dev list).

Anyway - I am not sure why we would go to the effort to replace one library with another **and** shade it, the point of shading is so you can have multiple versions of the library.  In other words why don't we just bump Guava here to the latest and shade that? is this just because google/guava#4011 has no fix at all (which is a perfectly acceptable answer - but then what about Jenkins core?)

Or not shade and make sure we are not using any of the known API breakages in Guava.  (I have a list I need to share that too!).
(`Itterators`, and `LoadingCache.getUnchecked` -> `LoadingCache.get` here from a glance at what has changed)

Anyway - Jenkins needs to be able to have library dependencies that are not exposed to plugins, and that is what I will be looking at shortly

Having said all this - the POM really should not expose the original non shaded libraries - that means plugins and the like will continue to get auto completion for things and more deps will sneak into plugins that really required, and causes things that will work at build time and blow up on deployment.

jglick:
Given the level of risk with _either_ Guava or Caffeine here, can we just get rid of both? For the two cache cases that are not bogus, use `[Concurrent]HashMap<KeyType, SoftReference<ValueType>>` and replace the `computeIfAbsent` with a couple lines of logic to check for either a null map value or a null referent. A bit more verbose but way less technical debt than the alternatives IMO.

jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
basil: It _is_ directly needed to avoid a compilation failure:

```
[ERROR] stapler/stapler/jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java:[68,36] variable tl is already defined in method getTagLibrary(java.lang.String)
```

core/src/main/java/org/kohsuke/stapler/Function.java
basil: > I doubt this ever worked to begin with

This would not have worked in the case of a class with a `fromStapler` method, because then the value would be a `MethodFunction` whose `Method m` has a `Class<?> clazz` that refers to the key of the map.

But it would have worked in the case of a class _without_ a `fromStapler` method, because then the value would be a `StaticFunction` whose `Method m` has a `Class<?> clazz` that refers to `org.kohsuke.stapler.Function` rather than the key of the map.

ben-manes:
Caffeine 2.x and 3.x are nearly identical, except for removal of deprecations. The major bump was to increment the JDK for removing Unsafe and some very minor API changes (e.g. returning a future for LoadingCache.refresh instead of void). If you use the latest 2.x and avoid deprecations, it should be forward compatible.

PR #214, opened 2021-04-22
[JENKINS-65442] json api improve performance when encoding unicode characters

scddev:
Improve performance when encoding escape sequences.

String.format has to compile the pattern for each character which consumes a lot of CPU time when requesting larger json data.

Use bit shift and lookup to simplify conversion.

Doing a small comparison on my notebook reveals:
* oldStyle 522ms

```
  public StringBuilder oldStyle() {
    StringBuilder b = new StringBuilder();
    for(int i=0;i<0xffff;i++) {
      b.append("\\u" + String.format("%04x", i));
    }
    return b;
  }
```

* newStyle 13ms
```
  private static final char[] HEX = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};

  public StringBuilder newStyle() {
    StringBuilder b = new StringBuilder();
    for(int i=0;i<0xffff;i++) {
      b.append("\\u");
      b.append(HEX[(i >> 12) & 0xf]);
      b.append(HEX[(i >> 8) & 0xf]);
      b.append(HEX[(i >> 4) & 0xf]);
      b.append(HEX[i & 0xf]);
    }
    return b;
  }
```


oleg-nenashev (approved):
Would be nice to have a benchmark data for it, e.g. JMH.
No objections if it is merged as is. CC @timja @jglick @daniel-beck as core maintainers

jglick (approved):

PR #215, opened 2021-04-26
Migrate from Guava's cache to `ConcurrentHashMap` in `JellyClassLoaderTearOff`

basil:
A subset of #213 that is hopefully non-controversial and safe to land in the short term because it does not involve Caffeine or shading.

### DIFF
--- a/jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
+++ b/jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
@@ -23,9 +23,6 @@
 
 package org.kohsuke.stapler.jelly;
 
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import org.apache.commons.jelly.JellyContext;
 import org.apache.commons.jelly.JellyException;
 import org.apache.commons.jelly.TagLibrary;
@@ -35,6 +32,8 @@ import org.kohsuke.stapler.MetaClassLoader;
 
 import java.lang.ref.WeakReference;
 import java.net.URL;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * {@link MetaClassLoader} tear-off for Jelly support.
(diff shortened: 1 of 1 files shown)


jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
```diff
@@ -47,7 +46,7 @@
     /**
      * See {@link JellyClassTearOff#scripts} for why we use {@link WeakReference} here.
```
jglick, line 47: Seems to be dead?

jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
```diff
@@ -56,46 +55,43 @@ public JellyClassLoaderTearOff(MetaClassLoader owner) {
...
                     if(owner.parent!=null) {
                         // parent first
-                        TagLibrary tl = owner.parent.loadTearOff(JellyClassLoaderTearOff.class).getTagLibrary(nsUri);
-                        if(tl!=null)    return tl;
+                        TagLibrary taglib = owner.parent.loadTearOff(JellyClassLoaderTearOff.class).getTagLibrary(key);
+                        if(taglib!=null)    return taglib;
                     }
 
-                    String taglibBasePath = trimHeadSlash(nsUri);
+                    String taglibBasePath = trimHeadSlash(key);
```
jglick, line 72: I think you can revert these lines, just use `nsUri`, and simply ignore the `key` arg. (When the second arg to `computeIfAbsent` is an inline lambda, the lambda arg is basically redundant. I wish there were an overload taking `Supplier` rather than `Function`—would be more convenient.)

jglick (approved):

jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
basil: Done.

jelly/src/main/java/org/kohsuke/stapler/jelly/JellyClassLoaderTearOff.java
basil: Sure, but this seems to be a preëxisting issue that I am making neither better nor worse in this PR.

jglick (approved):

PR #216, opened 2021-04-26
Inline Guava's `Iterators#limit`

basil:
A subset of #213 that is hopefully non-controversial and safe to land in the short term because it does not involve Caffeine or shading.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/export/Iterators.java
+++ b/core/src/main/java/org/kohsuke/stapler/export/Iterators.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2007 The Guava Authors
+ *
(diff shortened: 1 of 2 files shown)


jglick (approved):
Seems OK though I am unsure of the policy for inlining APL code into what is otherwise a BSD-2-Clause module. I would be inclined to just rewrite this one usage as I suspect that given the Javadoc any competent Java programmer could reimplement it (and we presumably do not even need all of it, like `remove` and `IllegalArgumentException`, just the very basic functionality).

basil:
I really don't think this is an issue, but if someone wants to reimplement this then feel free.

jglick:
FYI @jtnord

timja:
As long as the license header is kept intact it would be ok I expect?

PR #220, opened 2021-05-07
Replacing Guava’s `CacheBuilder` with Java Platform equivalents

jglick:
As proposed in https://github.com/stapler/stapler/pull/213#issuecomment-826981987, this completely removed the Guava dependency without adding anything in its place. Does not really make the code noticeably more complex. Continues #215. @basil @jtnord


core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -226,8 +222,9 @@ Object bindAndInvoke(Object o, StaplerRequest req, StaplerResponse rsp, Object..
             throw new AssertionError(e);    // impossible
         }
 
-        PARSE_METHODS = CacheBuilder.newBuilder().weakKeys().build(new CacheLoader<Class,Function>() {
-            public Function load(Class from) {
+        PARSE_METHODS = new ClassValue<Function>() {
+            @Override
+            public Function computeValue(Class<?> from) {
```
jglick, line 227: Not only simpler, but should fix a memory leak in the rare case that `fromStapler` is actually defined: https://github.com/stapler/stapler/pull/213#discussion_r620275869

jglick:
Lightly tested: ran Jenkins with this version of Stapler, did various GUI stuff, watched

```bash
java -jar …/jenkins-cli.jar -s http://localhost:8080/jenkins/ tail-log -l FINE org.kohsuke.stapler.AbstractTearOff
```

basil (approved):

basil:
Testing the incremental in jenkinsci/jenkins#5422. If it passes and this PR is merged and released, I'll mark that PR as ready to review so we can pick up this change from core.

PR #223, opened 2021-05-19
Bump various dependencies to latest versions

basil:
Bumps various dependencies to the latest versions. Jenkins core was already using the new versions from this PR at runtime, and Jenkins core is the only real consumer of Stapler, so this PR is more or less a no-op.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -49,7 +49,7 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>2.7</version>
+      <version>2.8.0</version>
     </dependency>
     <dependency>
       <groupId>commons-logging</groupId>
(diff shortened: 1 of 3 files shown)


core/pom.xml
```diff
@@ -84,7 +84,7 @@
        -->
       <groupId>org.codehaus.groovy</groupId>
       <artifactId>groovy-all</artifactId>
-      <version>1.8.3</version>
+      <version>2.4.12</version>
```
jglick, line 87: Best shared with the `stapler-groovy` module using `dependencyManagement`.

jglick (approved):

core/pom.xml
basil: Sure, I'll try and tackle this in some future cleanup PR.

jglick: That would be nice. I cannot recall offhand if Dependabot is clever enough to file a single PR with both bumps. (I know it does not work in `archetypes` but that is a weird case because the archetype POMs are not technically part of the reactor, and so they have to be added explicitly in Dependabot config sections.)

PR #226, opened 2021-05-24
Remove ASM dep from `ClassDescriptor`

jglick:
See https://github.com/stapler/stapler/pull/225#pullrequestreview-667138790. Unfortunately it does not seem to work consistently—the test passes using my IDE’s compile-on-save mode, but not via `mvn clean test`. If I am following correctly, the `LocalVariableTable` always includes the parameter names, which are then redundantly included in a `MethodParameters` section only when compiled with debug information? Need to look into why this would be. I suspect the difference is between bytecode and runtime retention.

### DIFF
--- a/core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
@@ -3,19 +3,14 @@ package org.kohsuke.stapler;
 import java.io.IOException;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.servlet.ServletException;
 import static org.junit.Assert.*;
 import org.junit.Test;
-import org.kohsuke.stapler.verb.GET;
-import org.kohsuke.stapler.verb.POST;
 
 /**
  * @author Alan Harder
@@ -28,11 +23,7 @@ public class ClassDescriptorTest {
         assertEquals("[a, b, x]",Arrays.asList(names).toString());
     }
 
-    @Test public void loadParametersFromAsm() throws Exception {
-        // get private method that is being tested
-        Method lpfa = ClassDescriptor.ASM.class.getDeclaredMethod(
-                "loadParametersFromAsm", Method.class);
-        lpfa.setAccessible(true);
+    @Test public void loadParameterNamesFromReflection() throws Exception {
         // collect test cases
         Map<String,Method> testCases = new HashMap<String,Method>();
         for (Method m : ClassDescriptorTest.class.getDeclaredMethods())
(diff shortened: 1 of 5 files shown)


jglick:
I think the problem is that `plugin-pom` does not default to setting https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#parameters so this information is unavailable. We could add this going forward, but then existing plugins would lack this information, and it could be years before they are all updated.

In some cases—web methods?—methods have associated `*.stapler` files which also contain the parameter names, but this does not apply to constructors and I do not think we can rely on it.

jglick:
Or there _is_ a `*.stapler` file for constructors? Sometimes?

`DispatcherTest.testInheritance` also fails: `org/kohsuke/stapler/DispatcherTest/Inheritance/doBar.stapler` is present but there is nothing for `Inheritance2`.

jglick:
Confirmed, the following makes tests pass:

```diff
diff --git core/pom.xml core/pom.xml
index c03db644e..4dcd86d9a 100644
--- core/pom.xml
+++ core/pom.xml
@@ -23,6 +23,24 @@
         </executions>
       </plugin>
     </plugins>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-compiler-plugin</artifactId>
+          <executions>
+            <execution>
+              <goals>
+                <goal>testCompile</goal>
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <parameters>true</parameters>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
   </build>
   <dependencies>
     <dependency>
```

jglick:
Could probably be reworked to try `MethodParameters` first (the most efficient and supported system), then fall back to using ASM for `LocalVariableTable` if necessary.

timja:
Interesting, if you're using spring boot, their maven / gradle plugins automatically add that command line flag these days.

I've been using the flag for users for simplifying models that use jackson as you don't need to duplicate the field name in the annotation as it can read it from the constructor parameter name.

Shouldn't it be in `compile` as well? not just `testCompile`?

jglick:
> Shouldn't it be in `compile` as well? not just `testCompile`?

If we are talking about `plugin-pom` (and `pom`), then yes of course. I was just pointing out that for purposes of unit tests in Stapler it suffices to add that POM snippet to make tests pass from Maven.

Ideally the annotation processor that writes `*.stapler` files could be amended to skip this if it can detect that method parameters are being written, to avoid redundancy, assuming these files are used for nothing else (I would have to research it a bit).

core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
```diff
@@ -48,8 +39,8 @@
         for (Map.Entry<String,String[]> entry : expected.entrySet()) {
             Method testMethod = testCases.get(entry.getKey());
             assertNotNull("Method missing for " + entry.getKey(), testMethod);
-            String[] result = (String[])lpfa.invoke(null, testMethod);
-            assertNotNull("Null result for " + entry.getKey());
+            String[] result = ClassDescriptor.loadParameterNamesFromReflection(testMethod);
+            assertNotNull("Null result for " + entry.getKey(), result);
```
jglick, line 43: Note that this fixes a test bug: this line was previously doing nothing, and a null `result` just led to an opaque NPE on the next line.

PR #233, opened 2021-05-30
Reduce ASM dep in `ClassDescriptor`

basil:
See [stapler/stapler#226 (comment)](https://github.com/stapler/stapler/pull/226#issuecomment-847359651). Amends that PR to try `MethodParameters` first (the most efficient and supported system) then fall back to using ASM for `LocalVariableTable` if necessary. Tested interactively in Jenkins core in the following scenarios:

- Stapler compiled with metadata for reflection on method parameters, Jenkins core compiled _without_ metadata for reflection on method parameters
- Both Stapler and Jenkins core compiled with metadata for reflection on method parameters via jenkinsci/pom#171

In the first scenario I verified that ASM was correctly used as a fallback. In the second scenario I verified that ASM was not used as a fallback.

### DIFF
--- a/core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
@@ -3,19 +3,14 @@ package org.kohsuke.stapler;
 import java.io.IOException;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.servlet.ServletException;
 import static org.junit.Assert.*;
 import org.junit.Test;
-import org.kohsuke.stapler.verb.GET;
-import org.kohsuke.stapler.verb.POST;
 
 /**
  * @author Alan Harder
@@ -28,11 +23,7 @@ public class ClassDescriptorTest {
         assertEquals("[a, b, x]",Arrays.asList(names).toString());
     }
 
-    @Test public void loadParametersFromAsm() throws Exception {
-        // get private method that is being tested
-        Method lpfa = ClassDescriptor.ASM.class.getDeclaredMethod(
-                "loadParametersFromAsm", Method.class);
-        lpfa.setAccessible(true);
+    @Test public void loadParameterNamesFromReflection() throws Exception {
         // collect test cases
         Map<String,Method> testCases = new HashMap<String,Method>();
         for (Method m : ClassDescriptorTest.class.getDeclaredMethods())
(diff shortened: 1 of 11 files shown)


core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
```diff
@@ -28,6 +28,32 @@
...
+        expected.put("NoParams_static", new String[0]);
+        expected.put("ManyParams", new String[] { "a", "b", "c", "d", "e", "f", "g", "h", "i" });
+        expected.put("Params_static", new String[] { "abc", "def", "ghi" });
+        // run tests
+        for (Map.Entry<String,String[]> entry : expected.entrySet()) {
+            Method testMethod = testCases.get(entry.getKey());
+            assertNotNull("Method missing for " + entry.getKey(), testMethod);
+            String[] result = ClassDescriptor.loadParameterNamesFromReflection(testMethod);
+            assertNotNull("Null result for " + entry.getKey(), result);
+            if (!Arrays.equals(entry.getValue(), result)) {
```
jglick, line 49: Could use a standard `Matcher` for this I think.

pom.xml
```diff
@@ -60,6 +60,9 @@
     <changelist>999999-SNAPSHOT</changelist>
     <scmTag>HEAD</scmTag>
     <spotbugs.excludeFilterFile>${project.basedir}/../src/spotbugs/spotbugs-excludes.xml</spotbugs.excludeFilterFile>
+
+    <!-- Generate metadata for reflection on method parameters -->
+    <maven.compiler.parameters>true</maven.compiler.parameters>
```
jglick, line 65: Better style to use plugin `configuration`.

jglick (approved):

jglick:
LMK if you intend to do any follow-ups; otherwise this can be merged as is.

pom.xml
basil: > Better style

Do you have a reference to a style guide?

core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
basil: I was just trying to change the existing code as little as possible. This is pretty much a copy of `loadParametersFromAsm`. There is a lot that can be cleaned up in this repository. Once the dependencies are up-to-date I'll start looking into cleanups like this.

pom.xml
jglick: Could not find one. I wrote up what I know: https://stackoverflow.com/a/67794133/12916

basil: OK, this still seems like a fairly subjective argument to me, but I trust your judgment.

core/src/test/java/org/kohsuke/stapler/ClassDescriptorTest.java
jglick: In particular the use of reflection in the ASM test is unnecessary (fixed in #226).

pom.xml
```diff
@@ -89,6 +89,21 @@
...
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-compiler-plugin</artifactId>
+          <executions>
+            <execution>
+              <goals>
+                <goal>compile</goal>
+              </goals>
+            </execution>
+          </executions>
```
jglick, line 101: No need to override this aspect of the superpom.

```suggestion
```

basil: Thanks for the suggestion. It should be obvious that I have very little knowledge, let alone understanding, of Maven.

pom.xml
```diff
@@ -89,6 +89,14 @@
             </ignores>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
```
jglick, line 93: By the way, for official plugins this can be omitted if you like:
```suggestion
```
or kept for legibility.

jglick (approved):

pom.xml
```diff
@@ -89,6 +89,14 @@
...
+          <artifactId>maven-compiler-plugin</artifactId>
+          <configuration>
+            <!-- Generate metadata for reflection on method parameters -->
+            <parameters>true</parameters>
+          </configuration>
+        </plugin>
```
jglick, line 99:
```suggestion
```

after #236

jglick:
Hmm https://github.com/stapler/stapler/pull/233/checks?check_run_id=2721994088 did not work?

PR #244, opened 2021-06-10
Revert "Stop using repackaged ASM; upgrade ASM to 9.1"

jtnord:
Reverts stapler/stapler#225

ASM does not eveolve in a compatable way and we do not want a moving ASM exposed from Jenkins core.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -88,9 +88,9 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <groupId>org.ow2.asm</groupId>
-      <artifactId>asm</artifactId>
-      <version>9.1</version>
+      <groupId>org.kohsuke</groupId>
+      <artifactId>asm5</artifactId>
+      <version>5.0.1</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci</groupId>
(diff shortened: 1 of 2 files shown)


bitwiseman:
@jtnord  I don't think this will address the problem.  
From the description of #225 : 

>So the upstream version of ASM has become part of the Jenkins API whether we like it or not. And so far this has not caused too much trouble, other than breaking Token Macro (via Parboiled). That issue was resolved in jenkinsci/token-macro-plugin#70, which updated Token Macro's POM to read:

```
    <dependency>
      <groupId>org.parboiled</groupId>
      <artifactId>parboiled-java</artifactId>
      <version>1.3.1</version>
      <exclusions>
        <!-- asm is provided by Jenkins core -->
        <exclusion>
          <groupId>org.ow2.asm</groupId>
          <artifactId>*</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

Reverting this change will not fix the problem.

jtnord:
Correct  but it prevent it worsening when we need to upgrade ASM to another
new version without having a bug alignment issue in plugins.

--
Sent from my phone, please excuse the brevity, typos and auto-correct
issues.

jglick (changes requested):
I do not think this solves anything, since Jenkins core was already exposing 9, and there is no risk of version skew between core and its bundled Stapler. Whether we should introduce _new_ dependencies from _plugins_ to the implicit version from core is another question.

jglick:
(FTR: discussion about next steps moved to dev list.)

PR #248, opened 2021-08-03
Escape control characters

basil:
See [Valid characters in XML](https://en.wikipedia.org/wiki/Valid_characters_in_XML). These are usually a sign that something else is going wrong, but unprintable characters are exceedingly difficult to debug. Escaping them at least makes them more visible and aids in diagnosis.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Stapler.java
+++ b/core/src/main/java/org/kohsuke/stapler/Stapler.java
@@ -1185,10 +1185,30 @@ public class Stapler extends HttpServlet {
             if(ch=='&')
                 buf.append("&amp;");
             else
+            if (isControlCharacter(ch)) {
+                buf.append("&#");
+                buf.append(Integer.toString(ch, 10));
+                buf.append(';');
+            } else
                 buf.append(ch);
         }
         if (buf.length()==v.length())   return  v;  // unmodified
         return buf.toString();
+    }
+
+    /**
+     * Determine whether the character is a control character. Such control characters are only
+     * valid in certain contexts in XML 1.1 documents, and their usage is restricted and highly
+     * discouraged.
+     *
+     * @param c The character to check.
+     * @return Whether the character is a control character.
+     */
+    private static boolean isControlCharacter(char c) {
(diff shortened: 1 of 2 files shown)


pzakha (approved):
Should we do anything if we encounter a NUL character?

basil:
> Should we do anything if we encounter a NUL character?

No. NUL characters are invalid in XML 1.1 and should not be escaped.

pzakha:
Right, I guess my question was: do we just pass them along and let the caller handle them or do we generate some sort of Exception? We don't necessarily have to handle it here, just something that came to mind reading this.

KalleOlaviNiemitalo:
Does this require more changes elsewhere, to emit the XML 1.1 declaration?

basil:
> Does this require more changes elsewhere, to emit the XML 1.1 declaration?

Not sure.

basil:
I tried updating the tests in jenkinsci/jenkins#5654 to try and parse the resulting XML with a real XML parser. Doing so reveals that it still doesn't parse, even with the escaped control characters, because of the lack of an XML 1.1 prolog. I'll have to think about how to modify Stapler to add such a prolog in a backward-compatible way.

KalleOlaviNiemitalo:
Is it possible to add the prolog only when the response contains those characters? Because Microsoft's .NET still does not support XML 1.1 ([ParseXmlDeclaration throws if the version is not 1.0](https://github.com/dotnet/runtime/blob/2387892317672b404dfaa0b46405b632f1f7976f/src/libraries/System.Private.Xml/src/System/Xml/Core/XmlTextReaderImpl.cs#L3898-L3915)), I think unconditionally emitting the XML 1.1 declaration is likely to break some API users even if they do not have these characters.

basil:
Yes I agree that backward compatibility is a concern. Your idea sounds reasonable to me at face value, but I am not a maintainer of this repository.

jglick:
Why not just throw an exception for all these cases? Then the client will get a clear indication that there is weird data here and they should switch to JSON. Or if the value is actually valid for some reason, the maintainer of the `@Exported` method ought to consider deprecating it in favor of an alternative that uses, say, Base64 encoding.

basil:
I no longer have a reason to continue working on this.

PR #286, opened 2021-11-29
Miscellaneous code cleanup

basil:
Fixing various IntelliJ IDEA warnings


timja (approved):
I looked through the more interesting looking commits, checkstyle / spotless could be added in a follow up for some like un-used imports

PR #322, opened 2022-01-24
Fail fast when the frontend sends an invalid class name

basil:
**Untested** PR to return better error messages for users who are affected by tables-to-divs issues. In [JENKINS-67218](https://issues.jenkins.io/browse/JENKINS-67218) and [JENKINS-67550](https://issues.jenkins.io/browse/JENKINS-67550), a tables-to-divs issue caused the frontend to send a list of classes rather than a single class to the backend. This list eventually made its way to the class loader, resulting in confusing error messages. Users thought that the classloader couldn't load the classes because they weren't present, but the real reason class loading is failing is that we're passing a list rather than a single class to the class loader. This PR changes the code to fail fast in this situation and give the user an informative error message before the problem makes its way to lower layers.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -698,13 +698,20 @@ public class RequestImpl extends HttpServletRequestWrapper implements StaplerReq
                 if(l==null) {// single value conversion
                     try {
                         Class actualType = type;
+                        boolean isArray = false;
                         String className = null;
                         if(j.has("stapler-class")) {
+                            if (j.optJSONArray("stapler-class") != null) {
+                                isArray = true;
+                            }
                             // deprecated as of 2.4-jenkins-4 but left here for a while until we are sure nobody uses this
                             className = j.getString("stapler-class");
                             LOGGER.log(FINE, "stapler-class is deprecated in favor of $class: {0}", className);
                         }
                         if(j.has("$class")) {
+                            if (j.optJSONArray("$class") != null) {
+                                isArray = true;
+                            }
                             className = j.getString("$class");
                         }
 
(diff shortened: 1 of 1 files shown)


basil:
Does the core team have any thoughts on this? I think that failing fast is generally always preferable to make the error more clear to users. If there are concerns about the error being too specific to Jenkins or tables-to-divs, I can always remove the link from the error message and keep it more generic.

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -698,20 +698,35 @@ public Object convertJSON(Object o) {
...
                         }
 
                         if (className != null) {
                             // sub-type is specified in JSON.
                             // note that this can come from malicious clients, so we need to make sure we don't have security issues.
 
+                            if (isArray) {
+                                throw new IllegalArgumentException(
+                                        "The frontend sent an unexpected list of classes ("
+                                                + className
```
jglick, line 725: What will an example value be?

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -698,20 +698,35 @@ public Object convertJSON(Object o) {
...
+                                                + ") rather than an expected single class. See"
+                                                + " https://www.jenkins.io/doc/developer/views/table-to-div-migration/"
```
jglick, line 727: I am pretty sure tables-to-divs is not the only case where this problem arises, though it may be the most common one recently.

jglick (approved):
Makes sense I think.

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
basil: Please read [JENKINS-67550](https://issues.jenkins.io/browse/JENKINS-67550) (which I linked to in the PR description). It contains an example value seen in the wild:

> `["org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition","org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition"]`

timja (approved):
can always iterate it if we find other usecases

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
jglick: So we could also check, say,

```java
className.contains(",")
```
?

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
jglick: Oh definitely include the link. I just meant it might be appropriate to adjust the phrasing to explain that this is only one possible root cause.

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
basil: Correct

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
basil: I will keep an eye out for other causes in the future and adjust the message then if needed.

PR #357, opened 2022-05-06
Use `PropertyResourceBundle` for localization

timja:
Fixes https://github.com/jenkinsci/stapler/issues/356

Prior art:
https://github.com/jenkinsci/stapler/pull/105

This would allow us to switch to utf-8 encoding in the future for localization when Java 11 is the baseline.

I wasted far too much time confused on why my tests were failing, turns out my IDE automatically created the encoding in iso-8859-1, at least in IntelliJ this is configurable per project:
https://www.jetbrains.com/help/idea/properties-files.html#encoding


jglick:
While I appreciate the sentiment, as in https://github.com/jenkinsci/stapler/pull/105#pullrequestreview-276717115 I am not sure this is prudent. Whoever added the ability to load `*.properties` in UTF-8 encoding neglected to define a way whereby a file could designate itself unambiguously as Unicode, for example with a BOM or a special file extension. Text editors may thus assume ISO-8859-1 encoding despite the project default, as for example NetBeans does, and show a Unicode file as mojibake. At runtime, https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/PropertyResourceBundle.html says

> Constructing a `PropertyResourceBundle` instance from an `InputStream` requires that the input stream be encoded in UTF-8. By default, if a `MalformedInputException` or an `UnmappableCharacterException` occurs on reading the input stream, then the `PropertyResourceBundle` instance resets to the state before the exception, re-reads the input stream in `ISO-8859-1`, and continues reading.

This seems like a nice heuristic but since anything is valid ISO-8859-1 you can never be sure that it was meant to be UTF-8: we might load some old `*.properties` using Latin-1 characters which happen to be valid (but nonsensical) UTF-8, causing mojibake to be displayed on Java 11. Before committing to this technique I would advise scanning the corpus of `*.properties` in the @jenkinsci org looking for any file which is not ASCII and yet is valid UTF-8 by accident. We would have to switch to Unicode escapes in all such files, to ensure they are displayed properly in both Java 8 and 11; and could not switch them to raw UTF-8 until the plugin picked a core baseline requiring 11.

XML (#171) would be a safer choice, though it is awkward to edit. JSON would be fine, but does not support comments, which we need. YAML would solve that problem and be more familiar to people born after the 1970s, but require us to package a parser in Stapler, conflicting with the library wrapper plugin (unless we shaded it, which has its own problems).

basil:
> Before committing to this technique I would advise scanning the corpus of *.properties in the https://github.com/jenkinsci org looking for any file which is not ASCII and yet is valid UTF-8 by accident.

I had some leftover spaghetti code from the Guava upgrade project to do searches like this, so I did a corpus search for `.properties` files that were both valid UTF-8 and valid ISO-8859-1 across all Jenkins and CloudBees plugins. I only found matches in 3 plugins, all maintained by @ikedam (the author of #105) and all in Japanese `.properties` files:

```
WEB-INF/classes/hudson/plugins/doclinks/artifacts/ArtifactsDocLinksConfig/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/hudson/plugins/doclinks/artifacts/ArtifactsDocLinksPublisher/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/hudson/plugins/doclinks/artifacts/Messages_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/groovy_label_assignment/GroovyLabelAssignmentAction/summary_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/groovy_label_assignment/GroovyLabelAssignmentProperty/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/groovy_label_assignment/Messages_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/Messages_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/preferences/BuildPreference/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/preferences/BuildPreferenceJobProperty/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/preferences/BuildPreferenceNodeProperty/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/preferences/Messages_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/rules/BuildResultScoringRule/config_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/rules/Messages_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/ScoringLoadBalancer/global_ja.properties is valid UTF-8 and valid ISO-8859-1
WEB-INF/classes/jp/ikedam/jenkins/plugins/scoringloadbalancer/util/Messages_ja.properties is valid UTF-8 and valid ISO-8859-1
```

Here are the links to the plugins:

- [DocLinks](https://plugins.jenkins.io/doclinks/)
- [Groovy Label Assignment](https://plugins.jenkins.io/groovy-label-assignment/)
- [Scoring Load Balancer](https://plugins.jenkins.io/scoring-load-balancer/)

These plugins haven't had any substantive commits or releases since 2016, so I think we can deprecate them and proceed with this change.

basil:
Even the results reported above appear to be false positives, since they are using Unicode escapes but just happen to contain comments with the unescaped text. In short I do not think compatibility is a legitimate concern.

jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
```diff
@@ -153,14 +157,24 @@ protected Properties get(String key) {
...
+                } catch (IOException ignored) {}
+            }
+
+            Enumeration<String> keys = propertyResourceBundle.getKeys();
+            // TODO Java 9+ can use 'asIterator' and get rid of below collections conversion
+            List<String> keysAsSaneType = Collections.list(keys);
+
+            for (String localKey : keysAsSaneType) {
+                String value = propertyResourceBundle.getString(localKey);
+                props.put(localKey, value);
```
basil, line 177: Per [the Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html):

> Because `Properties` inherits from `Hashtable`, the `put` and `putAll` methods can be applied to a `Properties` object. Their use is strongly discouraged as they allow the caller to insert entries whose keys or values are not `String`s. The `setProperty` method should be used instead.

```suggestion
                props.setProperty(localKey, value);
```

jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
```diff
@@ -153,14 +157,24 @@ protected Properties get(String key) {
...
-                    in.close();
-                }
+                propertyResourceBundle = new PropertyResourceBundle(in);
             } catch (IOException e) {
                 throw new Error("Failed to load "+url,e);
+            } finally {
+                try {
+                    in.close();
+                } catch (IOException ignored) {}
+            }
```
basil, line 169: This code could be simplified and the error handling improved to propagate the exception by removing these ancient (probably no longer necessary) workarounds and using a modern try-with-resources and `UncheckedIOException`:

```java
// attempt to load
props = new Properties();
String url = baseName + key + ".properties";
PropertyResourceBundle propertyResourceBundle;
try (InputStream in = openStream(url)) {
    propertyResourceBundle = new PropertyResourceBundle(in);
} catch (IOException e) {
    throw new UncheckedIOException("Failed to load " + url, e);
}
```

A mildly riskier refactoring, but I think it is probably safe. No strong feeling either way, though. Not a blocking comment.

jelly/src/test/java/org/kohsuke/stapler/jelly/ResourceBundleTest.java
```diff
@@ -0,0 +1,63 @@
...
+        String helloWorld = resourceBundle.format(Locale.CANADA, "hello_world");
+
+        assertThat(helloWorld, is("Welcome, World!"));
+    }
+
+    @Test
+    public void formatLocaleNonISO_8859_1() {
+        String helloWorld = resourceBundle.format(Locale.CHINA, "hello_world");
+
+        assertThat(helloWorld, is("你好，世界！简化"));
```
basil, line 44: This is failing to compile on Windows:

```
[2022-05-06T20:55:24.497Z] [INFO] --- maven-compiler-plugin:3.10.1:testCompile (default-testCompile) @ stapler-jelly ---
[2022-05-06T20:55:24.497Z] [INFO] Changes detected - recompiling the module!
[2022-05-06T20:55:24.497Z] [WARNING] File encoding has not been set, using platform encoding Cp1252, i.e. build is platform dependent!
[2022-05-06T20:55:24.497Z] [INFO] Compiling 15 source files to C:\Jenkins\workspace\Core_stapler_PR-357\jelly\target\test-classes
[2022-05-06T20:55:24.497Z] [INFO] -------------------------------------------------------------
[2022-05-06T20:55:24.497Z] [ERROR] COMPILATION ERROR : 
[2022-05-06T20:55:24.497Z] [INFO] -------------------------------------------------------------
[2022-05-06T20:55:24.497Z] [ERROR] C:\Jenkins\workspace\Core_stapler_PR-357\jelly\src\test\java\org\kohsuke\stapler\jelly\ResourceBundleTest.java:[44,52] error: unmappable character (0x81) for encoding windows-1252
[2022-05-06T20:55:24.497Z] [ERROR] C:\Jenkins\workspace\Core_stapler_PR-357\jelly\src\test\java\org\kohsuke\stapler\jelly\ResourceBundleTest.java:[54,49] error: unmappable character (0x81) for encoding windows-1252
[2022-05-06T20:55:24.497Z] [INFO] 2 errors 
[2022-05-06T20:55:24.497Z] [INFO] -------------------------------------------------------------
```

I suspect we need to side-port [these lines](https://github.com/jenkinsci/plugin-pom/blob/fe006ce973196bd311aa53dcdaee9fccc26e221b/pom.xml#L57-L59=) from `plugin-pom` to `pom`.

basil (approved):
Very nice! As mentioned in the preceding comments, I do not think compatibility is a legitimate concern here.

jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
basil: (If you don't want to go with the above refactoring, you could always call `e.addSuppressed(ignored)` to tape on the ignored exception from the `close()` to the primary exception.)

ikedam:
> I only found matches in 3 plugins, all maintained by @ikedam (the author of https://github.com/jenkinsci/stapler/pull/105) and all in Japanese .properties files

UTF8 letters are only in comments (for better experience in review) and I believe it won't cause mojibake.
Anyway, we can ignore those plugins for this context.

jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
timja: Thanks, I had to make some minor adjustment as tests failed due to needing to skip missing locale specific files (due to our custom resource bundle like implementation :)

but I think this should be safe, will run it through a core PR to make sure though

timja:
> Text editors may thus assume ISO-8859-1 encoding despite the project default, as for example NetBeans does, and show a Unicode file as mojibake

ha well I've found the issue you've commented all over, https://bz.apache.org/netbeans/show_bug.cgi?id=75906
Seems somewhat 'fixed', you can mark individual files as utf-8 or select multiple and do it. Not as nice as the IntelliJ setting although it's more flexible to a degree.

---- 

Huge thanks to @basil for doing the scan of the org

jelly/src/test/java/org/kohsuke/stapler/jelly/ResourceBundleTest.java
timja: That did the trick

NotMyFault (approved):
Thanks very much for taking care of it, Tim!

basil:
I think this copypasta in the `localization-support-plugin` needs to be adjusted in the same way:

https://github.com/jenkinsci/localization-support-plugin/blob/master/src/main/java/io/jenkins/plugins/localization/support/stapler/ResourceBundleFactoryImpl.java#L70-L74=

timja:
done in https://github.com/jenkinsci/localization-support-plugin/pull/6

jelly/src/main/java/org/kohsuke/stapler/jelly/ResourceBundle.java
```diff
@@ -142,29 +148,25 @@ protected Properties get(String key) {
...
+
+            for (String localKey : keysAsSaneType) {
+                String value = propertyResourceBundle.getString(localKey);
+                props.setProperty(localKey, value);
             }
+
+        } catch (FileNotFoundException ignored) {
+            // we fall back to the default properties file if a locale file is missing
+        } catch (IOException e) {
+            throw new UncheckedIOException("Failed to load " + url, e);
```
jglick, line 166:
```suggestion
            throw new UncheckedIOException("Failed to load " + url + ": " + e, e);
```

jglick (approved):
If compatibility is not a major concern then OK, may as well do it this way.

My understanding is that when updating a plugin to a baseline requiring Java 11 then you may—but need not—switch `*.properties` to UTF-8.

It would be possible to add an `hpi:` mojo to check your bundles vs. your Java version and advise if you should switch, but I am not sure it would be worth the effort (and overhead on every build).

PR #370, opened 2022-06-16
Use `ClassValue` for caching

jglick:
Generally a good idea for avoiding memory leaks, though that may not be a real issue in Jenkins since the `ClassLoader`s involved are those of plugins, which would never be collected during the JVM session anyway.

Motivated by a CloudBees CI thread dump showing a large number of threads blocked like this:

```
"Handling HEAD /jenkins/instance-identity/ from … : Jetty (winstone)-…" …
    - waiting to lock <…> (a java.util.HashMap)
      owned by "Handling GET /jenkins/job/someItem/configure from … : Jetty (winstone)-… SomeStuff/configure.jelly" …
    at org.kohsuke.stapler.WebApp.getMetaClass(WebApp.java:227)
    at org.kohsuke.stapler.WebApp.getMetaClass(WebApp.java:244)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:762)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:694)
    at org.kohsuke.stapler.Stapler.service(Stapler.java:240)
    at …
```

all of them blocked by one HTTP handler thread

```
"Handling GET /jenkins/job/someItem/configure from … : Jetty (winstone)-… SomeStuff/configure.jelly" …
    - waiting to lock <…> (a hudson.ClassicPluginStrategy$AntClassLoader2)
      owned by "Computer.threadPoolForRemoting [#…]" …
    at java.lang.Class.forName0(Native Method)
    at …
    at java.lang.reflect.Method.getParameterAnnotations(Method.java:639)
    at org.kohsuke.stapler.Function$InstanceFunction.getParameterAnnotations(Function.java:423)
    at jenkins.security.stapler.DoActionFilter.keep(DoActionFilter.java:97)
    at jenkins.security.stapler.TypedFilter.isRoutableMethod(TypedFilter.java:131)
    at jenkins.security.stapler.TypedFilter.isSpecificClassStaplerRelevant(TypedFilter.java:98)
    at jenkins.security.stapler.TypedFilter.isStaplerRelevant(TypedFilter.java:65)
    at jenkins.security.stapler.TypedFilter.isStaplerRelevantCached(TypedFilter.java:57)
    at jenkins.security.stapler.TypedFilter.isClassAcceptable(TypedFilter.java:50)
    at jenkins.security.stapler.TypedFilter.keep(TypedFilter.java:267)
    at org.kohsuke.stapler.FunctionList.filter(FunctionList.java:48)
    at org.kohsuke.stapler.MetaClass.buildDispatchers(MetaClass.java:192)
    at org.kohsuke.stapler.MetaClass.<init>(MetaClass.java:98)
    at org.kohsuke.stapler.WebApp.getMetaClass(WebApp.java:230)
    at org.kohsuke.stapler.jelly.groovy.GroovyFacet.createRequestDispatcher(GroovyFacet.java:71)
    at org.kohsuke.stapler.RequestImpl.getView(RequestImpl.java:269)
    at org.kohsuke.stapler.RequestImpl.getView(RequestImpl.java:264)
    at hudson.model.Descriptor.getHelpFile(Descriptor.java:756)
    at hudson.model.Descriptor.getHelpFile(Descriptor.java:738)
    at …
    at org.apache.commons.jexl.ExpressionImpl.evaluate(ExpressionImpl.java:80)
    at hudson.ExpressionFactory2$JexlExpression.evaluate(ExpressionFactory2.java:74)
    at org.apache.commons.jelly.tags.core.CoreTagLibrary$3.run(CoreTagLibrary.java:134)
    at org.apache.commons.jelly.tags.core.CoreTagLibrary$1.run(CoreTagLibrary.java:98)
    at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
    at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
    at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
    at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
    at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
    at org.kohsuke.stapler.jelly.JellyViewScript.run(JellyViewScript.java:95)
    at org.kohsuke.stapler.jelly.IncludeTag.doTag(IncludeTag.java:171)
    at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
    at …
    at org.kohsuke.stapler.Facet$1.dispatch(Facet.java:240)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:766)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
    at org.kohsuke.stapler.MetaClass$4.doDispatch(MetaClass.java:281)
    at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:58)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:766)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
    at org.kohsuke.stapler.MetaClass$4.doDispatch(MetaClass.java:281)
    at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:58)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:766)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:694)
    at org.kohsuke.stapler.Stapler.service(Stapler.java:240)
    at …
```

While a contributing factor was the `Computer.threadPoolForRemoting` thread (not shown), blocked in class loading predating paralellism (introduced I think in https://github.com/jenkinsci/jenkins/pull/5698), this seems silly because some of the rendered threads are trying to load unrelated classes: `WebApp.classMap` was a bottleneck.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/AnnotationHandler.java
+++ b/core/src/main/java/org/kohsuke/stapler/AnnotationHandler.java
@@ -23,12 +23,11 @@
 
 package org.kohsuke.stapler;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.beanutils.Converter;
 
 import javax.servlet.ServletException;
 import java.lang.annotation.Annotation;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 /**
  * Handles stapler parameter annotations by determining what values to inject for a method call.
@@ -68,20 +67,6 @@ public abstract class AnnotationHandler<T extends Annotation> {
         for (Annotation a : annotations) {
             Class<? extends Annotation> at = a.annotationType();
             AnnotationHandler h = HANDLERS.get(at);
-            if (h==null) {
-                InjectedParameter ip = at.getAnnotation(InjectedParameter.class);
-                if (ip!=null) {
-                    try {
-                        h = ip.value().newInstance();
-                    } catch (InstantiationException | IllegalAccessException e) {
-                        throw new ServletException("Failed to instantiate parameter injector for "+at,e);
-                    }
-                } else {
-                    h = NOT_HANDLER;
-                }
-                AnnotationHandler prev = HANDLERS.putIfAbsent(at, h);
-                if (prev!=null) h=prev;
-            }
             if (h==NOT_HANDLER)
                 continue;
             return h.parse(request,a,targetType,parameterName);
@@ -90,7 +75,22 @@ public abstract class AnnotationHandler<T extends Annotation> {
         return null; // probably we should report an error
     }
 
-    private static final ConcurrentMap<Class<? extends Annotation>,AnnotationHandler> HANDLERS = new ConcurrentHashMap<>();
+    private static final ClassValue<AnnotationHandler> HANDLERS = new ClassValue<AnnotationHandler>() {
+        @SuppressFBWarnings(value = "THROWS_METHOD_THROWS_RUNTIMEEXCEPTION", justification = "shut up")
+        @Override
+        protected AnnotationHandler computeValue(Class<?> at) {
+            InjectedParameter ip = at.getAnnotation(InjectedParameter.class);
+            if (ip != null) {
+                try {
+                    return ip.value().newInstance();
+                } catch (InstantiationException | IllegalAccessException e) {
+                    throw new RuntimeException("Failed to instantiate parameter injector for " + at, e);
+                }
+            } else {
+                return NOT_HANDLER;
+            }
+        }
+    };
 
     private static final AnnotationHandler NOT_HANDLER = new AnnotationHandler() {
         @Override
--- a/core/src/main/java/org/kohsuke/stapler/ClassLoaderValue.java
+++ b/core/src/main/java/org/kohsuke/stapler/ClassLoaderValue.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2022, CloudBees, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided
+ * that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright notice, this list of
+ *       conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright notice, this list of
+ *       conditions and the following disclaimer in the documentation and/or other materials
+ *       provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS
+ * OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.kohsuke.stapler;
+
+import java.lang.reflect.Proxy;
+
+/**
+ * Like {@link ClassValue} but for a whole {@link ClassLoader}.
+ */
+abstract class ClassLoaderValue<T> {
+
+    private final ClassValue<T> storage = new ClassValue<T>() {
+        @Override
+        protected T computeValue(Class<?> type) {
+            return ClassLoaderValue.this.computeValue(type.getClassLoader());
+        }
+    };
+
+    protected abstract T computeValue(ClassLoader cl);
+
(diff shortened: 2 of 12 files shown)


timja (approved):
👍 should probably run core tests first?

core/src/main/java/org/kohsuke/stapler/WebApp.java
```diff
@@ -224,13 +224,19 @@ public MetaClass getMetaClass(Class c) {
...
-                mc = new MetaClass(this,c);
-                classMap.put(c,mc);
+        if (c.navigator == KlassNavigator.JAVA) {
+            if (classMap == null) {
+                classMap = new ClassValue<MetaClass>() {
+                    @Override
+                    protected MetaClass computeValue(Class<?> c) {
+                        return new MetaClass(WebApp.this, Klass.java(c));
+                    }
+                };
```
basil, line 234: Presumably we are not concerned about concurrent callers because this is just a cache? Could use [double-checked locking](https://errorprone.info/bugpattern/DoubleCheckedLocking) here to eliminate the possibility of concurrent callers getting different cached values, which I think is a nice property to have in theory but likely does not matter much in practice.

core/src/main/java/org/kohsuke/stapler/TearOffSupport.java
```diff
@@ -38,44 +36,47 @@
...
-        tearOffs = r;
+    public final <T> T loadTearOff(Class<T> t) {
+        return t.cast(tearOffs.get(t));
     }
+
+    /**
+     * @deprecated Unused?
+     */
+    @Deprecated
+    public <T> void setTearOff(Class<T> type, T instance) {}
```
basil, line 81: The question mark does not exactly inspire confidence. Perhaps throwing `UnsupportedOperationException` or even `AssertionError` here would be preferable, as it would lead to callers (if they still exist) failing fast rather than driving on in a potentially invalid state.

core/src/main/java/org/kohsuke/stapler/ClassLoaderValue.java
```diff
@@ -0,0 +1,60 @@
...
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.kohsuke.stapler;
+
+import java.lang.reflect.Proxy;
+
+/**
+ * Like {@link ClassValue} but for a whole {@link ClassLoader}.
+ */
```
basil, line 30: Very creative!

core/src/main/java/org/kohsuke/stapler/AnnotationHandler.java
```diff
@@ -90,7 +75,22 @@ static Object handle(StaplerRequest request, Annotation[] annotations, String pa
         return null; // probably we should report an error
     }
 
-    private static final ConcurrentMap<Class<? extends Annotation>,AnnotationHandler> HANDLERS = new ConcurrentHashMap<>();
+    private static final ClassValue<AnnotationHandler> HANDLERS = new ClassValue<AnnotationHandler>() {
+        @SuppressFBWarnings(value = "THROWS_METHOD_THROWS_RUNTIMEEXCEPTION", justification = "shut up")
```
basil, line 79: My sentiment exactly.

basil (approved):

core/src/main/java/org/kohsuke/stapler/MetaClassLoader.java
```diff
@@ -72,7 +64,12 @@ public static MetaClassLoader get(ClassLoader cl) {
      *
      * Note that this permanently holds a strong reference to its key, i.e. is a memory leak.
```
basil, line 65: Also this comment is now out-of-date and should be rephrased.

basil (approved):

core/src/main/java/org/kohsuke/stapler/TearOffSupport.java
jglick: Or just deleting the method. I could not find any outside callers.

core/src/main/java/org/kohsuke/stapler/WebApp.java
jglick: `ClassValue` claims to have its own concurrency control. I do not think it matters here anyway since while these objects are not immutable, they do not hold valuable state.

jglick:
> run core tests

Probably. I ran `jenkins.security.stapler.**.*Test` successfully.

core/src/main/java/org/kohsuke/stapler/WebApp.java
basil: Yeah, I know that `ClassValue` is thread-safe; my point was that two concurrent callers to `getMetaClass` could wind up creating two separate instances of `ClassValue` (the second one clobbering the first) and thus the first instance of `ClassValue` could compute one value for a given key while the second instance of `ClassValue` could compute another value for the same key. As I wrote originally, probably does not matter in practice, but double-checked locking would be a clean way to make this "tighter" in theory. No strong preference either way.

jglick: Oh I see, you are concerned about https://github.com/jenkinsci/stapler/blob/0c572f5f289004ad7fd725e8c89cbba827a2f70c/core/src/main/java/org/kohsuke/stapler/WebApp.java#L123, not `ClassValue` behavior per se. I think harmless but it should be easy enough to prevent that.

basil: Indeed. The version from https://errorprone.info/bugpattern/DoubleCheckedLocking avoids `synchronized` in the common case (getting the volatile field after it has already been initialized) but this is definitely good enough, no need to over-optimize.

basil:
Well for a few more hours anyway. I am planning to merge https://github.com/jenkinsci/jenkins/pull/6083 today, just need to write a PR description. You know how it goes, writing the code is always the easy part, writing the docs always the hard part.

basil:
I just merged it so we should be good to go for testing this change now.

jglick:
https://github.com/jenkinsci/jenkins/pull/6693 passed.

basil:
> Or just YOLO-release; at worst we amend some problem or revert altogether and cut another release, if a DB PR turns up issues.

Issues have turned up indeed: https://github.com/jenkinsci/ec2-plugin/pull/771

PR #385, opened 2022-07-16
[JENKINS-60866] Make st:bind tag and JavaScript proxy work without inline JS

daniel-beck:
See [JENKINS-60866](https://issues.jenkins.io/browse/JENKINS-60866).

Currently, `st:bind` generates inline JS. This doesn't work with CSP, so move that out.

Additionally, add a new `StaplerRequest#createJavaScriptProxyParameters` as replacement for `StaplerRequest#createJavaScriptProxy` that can be used without needing to `eval` the result.

Downstream PR: https://github.com/jenkinsci/jenkins/pull/6865


jelly/src/main/java/org/kohsuke/stapler/jelly/BindTag.java
```diff
@@ -70,28 +72,41 @@ public void doTag(XMLOutput out) throws JellyTagException {
...
             throw new JellyTagException(e);
         }
     }
 
+    private void writeScriptTag(XMLOutput out, @CheckForNull Bound bound) throws SAXException {
+        final AttributesImpl attributes = new AttributesImpl();
+        if (bound == null) {
+            attributes.addAttribute("", "src", "src", "", Bound.getProxyScriptUrl(varName, null));
+        } else {
+            attributes.addAttribute("", "src", "src", "", bound.getProxyScriptURL(varName));
```
daniel-beck, line 103: Inconsistent naming needs fixing.

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,33 +65,66 @@ public abstract class Bound implements HttpResponse {
...
      */
     public final String getProxyScript() {
-        StringBuilder buf = new StringBuilder("makeStaplerProxy('").append(getURL()).append("','").append(
-                WebApp.getCurrent().getCrumbIssuer().issueCrumb()
-        ).append("',[");
+        return getProxyScript(getURL(), getTarget().getClass());
+    }
+
+    /**
+     * Returns the URl to the proxy script for this {@link org.kohsuke.stapler.bind.Bound}
```
daniel-beck, line 72:
```suggestion
     * Returns the URL to the proxy script for this {@link org.kohsuke.stapler.bind.Bound}
```

core/src/main/java/org/kohsuke/stapler/StaplerRequest.java
```diff
@@ -515,6 +517,36 @@ public interface StaplerRequest extends HttpServletRequest {
      * a proxy on the client side.
      *
      * Short cut for {@code getBoundObjectTable().bind(toBeExported).getProxyScript()}
+     *
+     * @deprecated Use {@link #createJavaScriptProxyParameters(Object)} and invoke {@code makeStaplerProxy} yourself.
      */
+    @Deprecated(since = "TODO")
```
jglick, line 523: Nobody is going to care about the version; keep it simple.

```suggestion
    @Deprecated
```

jglick (commented):
Do not recall well enough how `st:bind` works to offer a meaningful review

daniel-beck:
Note to self: @Wadeck recently mentioned we wouldn't actually want dynamically generated JS files to allow checksums in the future. For that, we could add a third mode that emits a JSON object (or directly assigns it to a context variable). That could them be parsed and the proxy object assigned by a utility function.

Wadeck:
⚠️ Not reviewed the PR yet, but got an occurrence of something interesting when starting the CSP analysis for popular plugins.

In JUnit, there is: https://github.com/jenkinsci/junit-plugin/blob/97ee857f0d37cbb72b88004cac347ca3d0f00d5c/src/main/resources/hudson/tasks/junit/History/index.jelly#L38-L41

```
<script>var view =
   <st:bind value="${it}"/>
</script>
<script type="text/javascript" src="${resURL}/plugin/junit/history/history.js"/>
```

Which could be ideally replaced by something like:
```
<st:bindVar var="view" value="${it}" />
<script type="text/javascript" src="${resURL}/plugin/junit/history/history.js"/>
```
Which will generate an inline script, with a hash that will be added to the header. That could be a way to inject such dynamic code, without having any other variables inside.
Now in general, using st:bind and equivalent should be discouraged due to the strong binding between Java and JavaScript that is not desired. (I feel the same pain as when I saw PHP database query directly inside views :trollface: )

timja:
It's a pattern from warnings-ng / echarts
https://github.com/jenkinsci/warnings-ng-plugin/blob/670dfe709bd31faf073ea895d11b96f07e512130/plugin/src/main/resources/io/jenkins/plugins/analysis/core/model/IssuesDetail/index.jelly#L27-L29

daniel-beck:
> ⚠️ Not reviewed the PR yet, but got an occurrence of something interesting when starting the CSP analysis for popular plugins.
> 
> In JUnit, there is: https://github.com/jenkinsci/junit-plugin/blob/97ee857f0d37cbb72b88004cac347ca3d0f00d5c/src/main/resources/hudson/tasks/junit/History/index.jelly#L38-L41
> 
> ```
> <script>var view =
>    <st:bind value="${it}"/>
> </script>
> <script type="text/javascript" src="${resURL}/plugin/junit/history/history.js"/>
> ```
> 
> Which could be ideally replaced by something like:
> 
> ```
> <st:bindVar var="view" value="${it}" />
> <script type="text/javascript" src="${resURL}/plugin/junit/history/history.js"/>
> ```
> 
> Which will generate an inline script

That has always worked: https://github.com/jenkinsci/stapler/blob/6dd4392b5e78a934976616f3a91aaf88b708dc4d/jelly/src/main/java/org/kohsuke/stapler/jelly/BindTag.java#L46-L56

daniel-beck:
> Note to self: @Wadeck recently mentioned we wouldn't actually want dynamically generated JS files to allow checksums in the future. For that, we could add a third mode that emits a JSON object (or directly assigns it to a context variable). That could then be parsed and the proxy object assigned by a utility function.

Might not apply, since checksums are for inline JS only, not external JS files? We "just" need to ensure there's no potential for dangerous injection for the name?

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,33 +65,66 @@ public abstract class Bound implements HttpResponse {
...
+
+    /**
+     * Returns the URL to the proxy script for this {@link org.kohsuke.stapler.bind.Bound}
+     * @param variableName the variable to assign to the bound object
+     * @param bound the bound object, or {@code null} if none.
+     * @return
+     */
+    public static String getProxyScriptUrl(String variableName, Bound bound) {
+        if (bound == null) {
+            // TODO Should this hardocde a non-UUID to ensure null bind?
```
jglick, line 79:
```suggestion
            // TODO Should this hardcode a non-UUID to ensure null bind?
```

jelly/src/main/java/org/kohsuke/stapler/jelly/BindTag.java
jglick: Still a consideration? ^^^

jglick (commented):
Not sure I follow well enough to review seriously.

daniel-beck:
As is, this change is broken for `WithWellKnownURL`. There's currently no implementation of that interface in the plugin ecosystem, but it seems to be the only viable workaround for [JENKINS-16341](https://issues.jenkins.io/browse/JENKINS-16341), so removing support for it seems not the correct way forward. Will need to amend this a bit.

My current thinking is that `#doScript` should, in the case of `WithWellKnownURL`, receive both the URL to the object, as well as the list of methods (for anything not `WithWellKnownURL`, the type and methods list can be looked up in the table). Alternatively, this could be done always.

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -50,6 +55,86 @@
...
+    }
+
+    private static boolean isValidJavaIdentifier(String name) {
+        if (name == null || StringUtils.isBlank(name)) {
+            return false;
+        }
+        if (!Character.isJavaIdentifierStart(name.charAt(0))) {
+            return false;
+        }
+        return name.substring(1).chars().allMatch(Character::isJavaIdentifierPart); // TODO More restrictive
```
daniel-beck, line 72: This should be more restrictive for security, as we're not escaping the method names for use in JS.

yaroslavafenkin: This seems fine the way it is. JS payloads would normally require symbols that are weird enough to not be valid Java identifiers.

yaroslavafenkin (approved):
Tested locally, seems to work as expected for cases I had in mind.

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -50,6 +55,88 @@
...
+        final String boundUrl = req.getRestOfPath();
+
+        if (var == null) {
+            return;
+        }
+
+        if (!isValidJavaScriptIdentifier(var)) {
+            LOGGER.log(Level.FINE, () -> "Rejecting invalid JavaScript identifier: " + var);
+            return;
+        }
```
daniel-beck, line 95: Breaks https://github.com/jenkinsci/build-monitor-plugin/blob/e07bbdfc6dac962ed030266aa556305f580100c0/build-monitor-plugin/src/main/resources/com/smartcodeltd/jenkinsci/plugins/buildmonitor/BuildMonitorView/index.jelly#L161

Current idea:
Change what the `st:bind` tag emits. Same (CSP-incompatible) behavior with no var name as with a complex/unsafe var name. That way callers with var names that are considered invalid still get their JS, just not from the new CSP-compatible path.

daniel-beck: https://github.com/jenkinsci/stapler/pull/385/commits/16f23e39438b731851dc446c81d32ec9c6c84589#diff-b2f036d8f4ef25d509293883bedcb5fba26848727b469d8c1d2ad4b314000203R90-R95

timja (approved):
/label ready-for-merge

----

This PR is now ready for merge, after ~24 hours, we will merge it if there's no negative feedback.

Thanks!

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,33 +65,77 @@ public abstract class Bound implements HttpResponse {
...
+        }
+    }
+
+    /**
+     * Returns the URL for the standalone proxy script.
+     * @param variableName the name of the JS variable to assign
+     * @return
+     */
+    public final String getProxyScriptURL(String variableName) {
+        final String methodsList = StringUtils.join(getBoundJavaScriptUrlNames(getTarget().getClass()).toArray(String[]::new), ",");
```
jglick, line 91: BTW with suitable imports
```suggestion
        final String methodsList = getBoundJavaScriptUrlNames(getTarget().getClass()).stream().collect(Collectors.joining(","));
```

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,33 +65,77 @@ public abstract class Bound implements HttpResponse {
...
+     *
+     * @param url URL to proxied object
+     * @param methods list of method names
+     * @return
+     */
+    public static String getProxyScript(String url, String[] methods) {
+        StringBuilder buf = new StringBuilder("makeStaplerProxy('").append(url).append("','").append(
+                WebApp.getCurrent().getCrumbIssuer().issueCrumb()
+        ).append("',[").append(
+                StringUtils.join(Arrays.stream(methods).sorted().map(it -> "'" + it + "'").toArray(), ",")).append("])");
```
jglick, line 138:
```suggestion
                Arrays.stream(methods).sorted().map(it -> "'" + it + "'").collect(Collectors.joining(","))).append("])");
```

jglick (commented):
Presuming someone else has done code review & testing on this & downstream, do you need a merge & release? It is not clear to me whether this is still a WiP or ready to go.

daniel-beck:
> do you need a merge & release? It is not clear to me whether this is still a WiP or ready to go.

Thanks, just no good time IMO in the recent weeks. I'd like to be around for post-release support and ideally not simultaneously with another Stapler release.

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
daniel-beck: Could be simplified further using `String#join`.

daniel-beck:
Last call for reviews :)

timja (approved):

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,34 +65,76 @@ public abstract class Bound implements HttpResponse {
...
-                WebApp.getCurrent().getCrumbIssuer().issueCrumb()
-        ).append("',[");
+        return getProxyScript(getURL(), getTarget().getClass());
+    }
+
+    /**
+     * Returns the URL to the proxy script for this {@link org.kohsuke.stapler.bind.Bound}
+     * @param variableName the variable to assign to the bound object
+     * @param bound the bound object, or {@code null} if none.
+     * @return
```
jglick, line 75: delete, or fill in

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,34 +65,76 @@ public abstract class Bound implements HttpResponse {
...
+            return Stapler.getCurrentRequest().getContextPath() + BoundObjectTable.SCRIPT_PREFIX + "/null?var=" + variableName;
+        } else {
+            return bound.getProxyScriptURL(variableName);
+        }
+    }
+
+    /**
+     * Returns the URL for the standalone proxy script.
+     * @param variableName the name of the JS variable to assign
+     * @return
```
jglick, line 88: ditto

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
```diff
@@ -63,34 +65,76 @@ public abstract class Bound implements HttpResponse {
...
+    /**
+     * Returns the Stapler proxy script for the specified URL and method names
+     *
+     * @param url URL to proxied object
+     * @param methods list of method names
+     * @return
+     */
+    public static String getProxyScript(String url, String[] methods) {
+        final String crumb = WebApp.getCurrent().getCrumbIssuer().issueCrumb();
+        final String methodNamesList = Arrays.stream(methods).sorted().map(it -> "'" + it + "'").collect(Collectors.joining(","));
```
jglick, line 136: or
```suggestion
        final String methodNamesList = Arrays.stream(methods).sorted().collect(Collectors.joining("','", "'", "'"));
```

jelly/src/main/java/org/kohsuke/stapler/jelly/BindTag.java
```diff
@@ -68,25 +71,54 @@ public void doTag(XMLOutput out) throws JellyTagException {
...
+                    writeScriptTag(out, h);
+                }
             }
         } catch (SAXException e) {
             throw new JellyTagException(e);
         }
     }
+
+    /**
+     * Writes a &lt;script&gt; tag whose {@code src} attribute points to
```
jglick, line 107:
```suggestion
     * Writes a {@code <script>} tag whose {@code src} attribute points to
```

jglick (commented):
Superficial comments, since I do not remember much about this system.

core/src/main/java/org/kohsuke/stapler/bind/Bound.java
daniel-beck: I specifically decided against that as it's not a standalone block of sorts -- the end of the quote for the crumb would be included, or the close paren, which aren't specific to this list. Hence the slight code reorg to make the concatenation more readable.

PR #394, opened 2022-08-07
Migrate from Hickory to Elementary

basil:
We currently test our annotation processors with [Hickory](https://github.com/vietj/hickory), an outdated library that hasn't been maintained in over a decade. This PR switches us to [Elementary](https://github.com/Pante/elementary), a modern and well-maintained annotation processor testing library. Note that Elementary requires both Java 11 and JUnit 5, both of which we now support.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,6 +120,10 @@
       <id>repo.jenkins-ci.org</id>
       <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
+    <repository>
+      <id>elementary-releases</id>
+      <url>https://repo.karuslabs.com/repository/elementary-releases/</url>
+    </repository>
   </repositories>
 
   <pluginRepositories>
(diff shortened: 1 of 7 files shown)


MarkEWaite (approved):

timja (approved):

pom.xml
```diff
@@ -120,6 +120,10 @@
       <id>repo.jenkins-ci.org</id>
       <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
+    <repository>
+      <id>elementary-releases</id>
+      <url>https://repo.karuslabs.com/repository/elementary-releases/</url>
+    </repository>
```
jglick, line 126: This is only pending https://github.com/jenkins-infra/helpdesk/issues/3115 right?

daniel-beck:
> well-maintained

Actively perhaps, but well? https://github.com/Pante/elementary/issues/153#issuecomment-1280182769 should just not happen.

PR #413, opened 2022-10-26
Some exception handling that was forgotten as part of JEP-227

Vlatombe:
Existing logic introduced in https://github.com/jenkinsci/stapler/commit/1a07292090c93748eccdf1864a4e327941db8fc6

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Stapler.java
+++ b/core/src/main/java/org/kohsuke/stapler/Stapler.java
@@ -98,10 +98,9 @@ public class Stapler extends HttpServlet {
 
     /**
      * Exceptions we don't want to print a stacktrace for because they are normal behaviour.
-     * [HUDSON-4834] A stack trace is too noisy for this; could just need to log in.
-     * (Could consider doing this for all AcegiSecurityException's.)
+     * [JENKINS-4834] A stack trace is too noisy for this; could just need to log in.
      */
-    private  static final List<String> EXCEPTIONS_DO_NOT_PRINT_STACKTRACE = Arrays.asList(
+    private  static final Set<String> EXCEPTIONS_DO_NOT_PRINT_STACKTRACE = Set.of(
             "org.acegisecurity.AccessDeniedException",
             "org.springframework.security.access.AccessDeniedException"
     );
--- a/core/src/main/java/org/kohsuke/stapler/Stapler.java
+++ b/core/src/main/java/org/kohsuke/stapler/Stapler.java
@@ -96,6 +96,15 @@ import static org.kohsuke.stapler.Dispatcher.*;
  */
 public class Stapler extends HttpServlet {
 
+    /**
+     * Exceptions we don't want to print a stacktrace for because they are normal behaviour.
+     * [HUDSON-4834] A stack trace is too noisy for this; could just need to log in.
+     * (Could consider doing this for all AcegiSecurityException's.)
+     */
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -802,9 +811,7 @@ boolean tryInvoke(RequestImpl req, ResponseImpl rsp, Object node ) throws IOExce
                     if (!isSocketException(e)) {
                         getServletContext().log("Error while serving " + url, e);
                     }
-                } else if (c.getName().equals("org.acegisecurity.AccessDeniedException")) {
-                    // [HUDSON-4834] A stack trace is too noisy for this; could just need to log in.
-                    // (Could consider doing this for all AcegiSecurityException's.)
+                } else if (EXCEPTIONS_DO_NOT_PRINT_STACKTRACE.contains(c.getName())) {
```
jglick, line 814: Still not enough I think because it does not account for subclasses, like `AccessDeniedException3` in Jenkins core. Would need to walk the inheritance stack. Also I think you could just switch to solely checking the Spring type now, since newer versions of Stapler will only be used with newer versions of Jenkins core.

I am not sure why Stapler is even logging anything here. It is already throwing `ServletException`; should be up to the container or various filters to decide what to do with it.

Vlatombe: It actually walk through super classes already

jglick: Oh right, missed that somehow.

core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -96,6 +96,15 @@
  */
 public class Stapler extends HttpServlet {
 
+    /**
+     * Exceptions we don't want to print a stacktrace for because they are normal behaviour.
+     * [HUDSON-4834] A stack trace is too noisy for this; could just need to log in.
```
jglick, line 101:
```suggestion
     * [JENKINS-4834] A stack trace is too noisy for this; could just need to log in.
```

https://issues.jenkins.io/browse/JENKINS-4834 FTR

core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -96,6 +96,15 @@
...
+     * (Could consider doing this for all AcegiSecurityException's.)
```
jglick, line 102:
```suggestion
```

since there is no Spring Security equivalent:

```java
public class AccessDeniedException extends RuntimeException {
```

core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -96,6 +96,15 @@
...
+     */
+    private  static final List<String> EXCEPTIONS_DO_NOT_PRINT_STACKTRACE = Arrays.asList(
```
jglick, line 104:
```suggestion
    private  static final Set<String> EXCEPTIONS_DO_NOT_PRINT_STACKTRACE = Set.of(
```

jglick (approved):

basil (approved):
Thanks for the PR!

PR #438, opened 2023-01-27
Delete unused Jelly view for `LargeText`

jglick:
Does not seem to have been touched since 2009, and AFAIK is unused: Jenkins core has its own version of this, used from `console.jelly`, modified for example in https://github.com/jenkinsci/jenkins/pull/6866.


core/src/main/resources/org/kohsuke/stapler/framework/io/LargeText.js
```diff
@@ -1,86 +0,0 @@
-//
-// Auto-scroll support for progressive log output.
-//   See http://radio.javaranch.com/pascarello/2006/08/17/1155837038219.html
-//
-// @include org.kohsuke.stapler.framework.prototype.prototype
```
jglick, line 5: Probably means we could also delete https://github.com/jenkinsci/stapler/blob/dcb187defb7e6880871d82578dfdf1a064a2fbf3/core/src/main/resources/org/kohsuke/stapler/framework/prototype/prototype.js#L1-L2

Vlatombe (approved):

timja (approved):

core/src/main/resources/org/kohsuke/stapler/framework/io/LargeText.js
jglick: Perhaps not: https://github.com/jenkinsci/stapler/blob/dcb187defb7e6880871d82578dfdf1a064a2fbf3/jelly/src/main/java/org/kohsuke/stapler/jelly/BindTag.java#L98-L106 (as well as a few other references floating around)

PR #452, opened 2023-04-03
Remove Stapler dependency on Prototype or jQuery

basil:
This PR eliminates Stapler's dependencies on any JavaScript frameworks in favor of the Fetch API. Previously it supported either jQuery or Prototype, although the jQuery code path was dead code and would never have worked with Jenkins anyway. This PR rips out all this code in favor of native JavaScript. Core will require a very minor adaptation to deal with this change; see https://github.com/jenkinsci/jenkins/pull/7805.

To test this I ran this repository's test and core's tests. I also did manual interactive testing of the Configure Project page and the Build History page, which exercised both of the use cases of `bind.js`: the JSON use case and the HTML use case. In each use case I stepped through the code before and after in the debugger to verify it was working the same way with this PR. More detail in the self-review. Once I get an incremental I plan on running this through BOM as well.


core/src/main/resources/org/kohsuke/stapler/bind.js
```diff
@@ -1,6 +1,6 @@
 // bind tag takes care of the dependency as an adjunct
 
-function makeStaplerProxy(url,crumb,methods) {
+function makeStaplerProxy(url,staplerCrumb,methods) {
```
basil, line 3: Renaming this to avoid a clash with the Jenkins global variable named `crumb`, which is needed to set the "Jenkins-Crumb" header later below (not to be confused with the "Crumb" header, which is what this second argument is for…).

core/src/main/resources/org/kohsuke/stapler/bind.js
```diff
@@ -26,39 +26,43 @@ function makeStaplerProxy(url,crumb,methods) {
             for (var i=0; i<args.length-(callback!=null?1:0); i++)
                 a.push(args[i]);
 
-            if(window.jQuery === window.$) { //Is jQuery the active framework?
```
basil, line 29: Let it be said that this dead jQuery code path would have never worked with Jenkins. `progressiveRendering.jelly` expects a JSON response and invokes `callback.responseObject()` to get the JSON. `renderOnDemand.jelly` expects HTML and invokes `callback.responseText` to get the HTML. This dead jQuery code path was only defining `responseObject`, so it would have worked with `progressiveRendering.jelly` but not `renderOnDemand.jelly`.

core/src/main/resources/org/kohsuke/stapler/bind.js
```diff
@@ -26,39 +26,43 @@ function makeStaplerProxy(url,crumb,methods) {
...
-                    success: function(data, textStatus, jqXHR) {
-                        if (callback!=null) {
+            var headers = {
+                'Content-Type': 'application/x-stapler-method-invocation;charset=UTF-8',
+                'Crumb': staplerCrumb,
+            };
+            // If running in Jenkins, add Jenkins-Crumb header.
+            if (typeof crumb !== 'undefined') {
+                headers = crumb.wrap(headers);
+            }
```
basil, line 36: In the Prototype version, this was done deep inside of the guts of `Ajax.Request`, which Kohsuke had patched to automatically set this header. Now that we aren't using Jenkins' patched Prototype, we have to do this ourselves. This is a leaky abstraction, but it is not the only place in Stapler where we have Jenkins-specific logic.

core/src/main/resources/org/kohsuke/stapler/bind.js
```diff
@@ -26,39 +26,43 @@ function makeStaplerProxy(url,crumb,methods) {
...
+
+            fetch(url + methodName, {
+                method: 'POST',
+                headers: headers,
+                body: stringify(a),
```
basil, line 41: Note: `JSON.stringify` _cannot_ be used here, because it blows up when Prototype is present (which is not the case in this repository's test suite, but _is_ the case in core's test suite and core's production environment). Fortunately Kohsuke did provide a custom `stringify` method for use with the dead jQuery code path, which we can use here and which solves the problem.

core/src/main/resources/org/kohsuke/stapler/bind.js
```diff
@@ -26,39 +26,43 @@ function makeStaplerProxy(url,crumb,methods) {
...
+            })
+            .then(function(response) {
+                if (response.ok) {
+                    if (response.headers.has('content-type') && response.headers.get('content-type').startsWith('application/json')) {
```
basil, line 45: The next few lines are a bit tricky, so let me break it down. `progressiveRendering.jelly` expects a JSON response and invokes `callback.responseObject()` to get the JSON. `renderOnDemand.jelly` expects HTML and invokes `callback.responseText` to get the HTML. So there are two supported use cases. This repository's tests only cover the JSON use case, so it is unclear whether the usage of the HTML usage in core was originally intended or if that usage happened to come about by exploiting the existing implementation. Nevertheless, that use case now exists, so it must continue to be supported.

The old code got the `t` object from Prototype, which had a `responseText` field. This was used for HTML use case previously. To accommodate the JSON use case, that `responseText` was put in parentheses and `eval`'d to define the `responseObject()` function. So in the old code, the object passed to the callback had both `responseText` and `responseObject()` defined. If you expected JSON and called `responseText` you'd get garbage, and if you expected text and called `responseObject()` you'd also get garbage. (Nobody actually did this.)

The new code is a lot cleaner and differentiates between the two use cases by checking the response header. If it's JSON, we parse the JSON with the Fetch API (not `eval`!) and return an object with only a `responseObject()` function. If it's text, we return a function with only `responseText`. Simple and straightforward.

core/src/main/resources/org/kohsuke/stapler/framework/prototype/prototype.js
```diff
@@ -1,4221 +0,0 @@
-/*  Prototype JavaScript framework, version 1.6.0.2
- *  (c) 2005-2008 Sam Stephenson
```
basil, line 2: Take a close look at the version number and those dates.

core/src/test/java/org/kohsuke/stapler/interceptor/JsonOutputFilterTest.java
```diff
@@ -117,16 +117,12 @@ public MyData getSomeIncludedData() {
     public void doIndex(StaplerResponse rsp) throws IOException {
         rsp.setContentType("text/html");
         PrintWriter w = rsp.getWriter();
-        w.println("<html><body><script src='prototype'></script><script src='script'></script>");
+        w.println("<html><body><script src='script'></script>");
         w.println("<script>var v = "+ WebApp.getCurrent().boundObjectTable.bind(this).getProxyScript()+";</script>");
-        w.println("<script>var callback = function(t){var x=t.responseObject(); alert(Object.toJSON(x)); };</script>");
+        w.println("<script>var callback = function(t){var x=t.responseObject(); alert(JSON.stringify(x)); };</script>");
```
basil, line 122: Since this test suite doesn't have Prototype present, it's OK to call `JSON.stringify()` here. (The test suite in core, and core itself in production, does have Prototype present, though.)

core/src/test/java/org/kohsuke/stapler/test/JettyTestCase.java
```diff
@@ -77,6 +81,24 @@ protected void configure(ServletContextHandler context) {
...
     }
 
+    /**
+     * Get the {@link WebClient} preconfigured for testing.
+     */
+    protected WebClient createWebClient() {
+        WebClient webClient = new WebClient();
+        webClient.getOptions().setFetchPolyfillEnabled(true);
+        return webClient;
+    }
```
basil, line 91: Needed for HtmlUnit to be able to invoke the Fetch API.

core/src/test/java/org/kohsuke/stapler/test/JettyTestCase.java
```diff
@@ -77,6 +81,24 @@ protected void configure(ServletContextHandler context) {
...
+
+    /**
+     * Blocks until the ENTER key is hit.
+     * This is useful during debugging a test so that one can inspect the state through the web browser.
+     */
+    protected void interactiveBreak() throws IOException {
+        System.out.println("Jetty is running at " + url);
+        new BufferedReader(new InputStreamReader(System.in, Charset.defaultCharset())).readLine();
+    }
```
basil, line 100: I discovered a need for this while debugging.

jelly/src/main/java/org/kohsuke/stapler/jelly/BindTag.java
```diff
@@ -91,40 +89,4 @@ public void doTag(XMLOutput out) throws JellyTagException {
...
     }
-
-    /**
-     * Ensures that we have the dependencies properly available to run bind.js.
-     *
-     * bind.js requires either prototype or jQuery. If we use jQuery, we use
-     * {@code JSON.stringify()}, which isn't present in older browser, so we
-     * will add it.
-     */
-    private void ensureDependencies(XMLOutput out) throws JellyTagException {
```
basil, line 102: There are no dependencies left to ensure.

jelly/src/test/java/org/kohsuke/stapler/jelly/BindTagTest.java
```diff
@@ -23,19 +24,19 @@ protected void setUp() throws Exception {
...
-        page.executeJavaScript("v.foo('hello world');");
+        page.executeJavaScript("v.foo('hello world', 2);");
         wc.getJavaScriptEngine().processPostponedActions();
         wc.waitForBackgroundJavaScript(10000);
         assertEquals("hello world",value);
+        assertEquals(2, number);
     }
 
-    public void jsFoo(String arg) {
+    public void jsFoo(String arg, int arg2) {
```
basil, line 38: This test was copied to a jQuery version, which was later enhanced by supporting an extra argument (probably to test some additional use case). I am now deleting the jQuery version of this test, but I thought I'd keep this test enhancement and backport it to the original test, since if someone thought it worthwhile to enhance the other test, that same logic probably applies here as well.

timja (approved):
Thanks! Great cleanup.

timja (approved):

basil:
Causes [JENKINS-71236](https://issues.jenkins.io/browse/JENKINS-71236).

PR #493, opened 2023-10-03
[JENKINS-72113] Relax serialization of `BoundObjectTable.Table`

Vlatombe:
[JENKINS-72113](https://issues.jenkins.io/browse/JENKINS-72113)

### Testing done

```[tasklist]
### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
```

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
+++ b/core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
@@ -35,6 +35,7 @@ import org.kohsuke.stapler.StaplerResponse;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpSession;
 import java.io.IOException;
+import java.io.Serializable;
 import java.lang.ref.WeakReference;
 import java.util.HashMap;
 import java.util.Map;
@@ -112,7 +113,7 @@ public class BoundObjectTable implements StaplerFallback {
     /**
      * Per-session table that remembers all the bound instances.
      */
-    public static class Table {
+    public static class Table implements Serializable {
         private final Map<String,Ref> entries = new HashMap<>();
         private boolean logging;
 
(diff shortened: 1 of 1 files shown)


jglick (approved):

daniel-beck (approved):
Noting that while this will skip some (most?) bound objects, possibly resulting in the current page being broken after session deserialization (requiring reload), that's a net improvement over not being able to serialize the session at all (and being logged out, having to log in again all over) in a suitably set up environment.

jglick:
Right, this brings Stapler closer to compliance with the [spec](https://jakarta.ee/specifications/servlet/5.0/jakarta-servlet-spec-5.0#distributed-environments) because at least other parts of the session, most crucially the login state, can be distributed.

To expand: in a deserialized session, https://github.com/jenkinsci/stapler/blob/892bddc2ea0a4e12a23e004c960c84f31a172aad/core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java#L172-L175 might be triggered. The impact would be that `/$stapler/bound/…` requests would give a 404 from https://github.com/jenkinsci/stapler/blob/892bddc2ea0a4e12a23e004c960c84f31a172aad/core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java#L157-L159 so any JavaScript running on the page (such as from a `l:progressiveRendering`) would stop with an error and the user would need to reload the page. (Similar to what already happens if your session expires while the page is open, or you log out in another tab, or the controller restarts, etc.)

This could be improved in the future to

* `st:bind` a `Serializable` object
* implement `WithWellKnownURL`
* use other mechanisms for dynamic page content (other JS frameworks, REST API calls)

on a case-by-case basis; certain usages of `st:bind` (such as `ProgressiveRendering`) do not seem to lend themselves to distributed state without an incompatible rewrite.

In a standard Winstone setup, sessions are never serialized so this patch would have no effect.

PR #542, opened 2024-05-10
[JENKINS-73170] Upgrade Commons FileUpload from 1.5 to 2.0.0-M2

basil:
See https://github.com/jenkinsci/jenkins/pull/9259 for the PR description.

### Testing done

See https://github.com/jenkinsci/jenkins/pull/9259.

### DIFF
--- a/core/src/main/java/org/apache/commons/fileupload/FileItem.java
+++ b/core/src/main/java/org/apache/commons/fileupload/FileItem.java
@@ -1,0 +1,438 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.fileupload;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+import java.nio.file.Path;
+import org.apache.commons.fileupload2.core.FileItemHeadersProvider;
+
+/**
+ * <p> This class represents a file or form item that was received within a
+ * <code>multipart/form-data</code> POST request.
+ *
+ * <p> After retrieving an instance of this class from a {@code
+ * org.apache.commons.fileupload.FileUpload FileUpload} instance (see
+ * {@code org.apache.commons.fileupload.servlet.ServletFileUpload
+ * #parseRequest(javax.servlet.http.HttpServletRequest)}), you may
+ * either request all contents of the file at once using {@link #get()} or
+ * request an {@link java.io.InputStream InputStream} with
+ * {@link #getInputStream()} and process the file without attempting to load
+ * it into memory, which may come handy with large files.
+ *
+ * <p> While this interface does not extend
+ * <code>javax.activation.DataSource</code> per se (to avoid a seldom used
+ * dependency), several of the defined methods are specifically defined with
+ * the same signatures as methods in that interface. This allows an
+ * implementation of this interface to also implement
+ * <code>javax.activation.DataSource</code> with minimal additional work.
+ */
+public interface FileItem {
+
+    // ------------------------------- Methods from javax.activation.DataSource
+
+    /**
+     * Returns an {@link java.io.InputStream InputStream} that can be
+     * used to retrieve the contents of the file.
+     *
+     * @return An {@link java.io.InputStream InputStream} that can be
+     *         used to retrieve the contents of the file.
+     *
+     * @throws IOException if an error occurs.
+     */
+    InputStream getInputStream() throws IOException;
+
+    /**
+     * Returns the content type passed by the browser or <code>null</code> if
+     * not defined.
+     *
+     * @return The content type passed by the browser or <code>null</code> if
+     *         not defined.
+     */
+    String getContentType();
+
+    /**
+     * Returns the original filename in the client's filesystem, as provided by
+     * the browser (or other client software). In most cases, this will be the
+     * base file name, without path information. However, some clients, such as
+     * the Opera browser, do include path information.
+     *
+     * @return The original filename in the client's filesystem.
+     * @throws InvalidFileNameException The file name contains a NUL character,
+     *   which might be an indicator of a security attack. If you intend to
+     *   use the file name anyways, catch the exception and use
+     *   InvalidFileNameException#getName().
+     */
+    String getName();
+
+    // ------------------------------------------------------- FileItem methods
+
+    /**
+     * Provides a hint as to whether or not the file contents will be read
+     * from memory.
+     *
+     * @return <code>true</code> if the file contents will be read from memory;
+     *         <code>false</code> otherwise.
+     */
+    boolean isInMemory();
+
+    /**
+     * Returns the size of the file item.
+     *
+     * @return The size of the file item, in bytes.
+     */
+    long getSize();
+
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/commons/fileupload/FileItem.java
```diff
@@ -0,0 +1,438 @@
...
+     * {@link FileItem} or {@code FileItemStream} should implement this
+     * interface to be able to get the raw headers found within the item
+     * header block.
+     *
+     * @param headers the instance that holds onto the headers
+     *         for this instance.
+     */
+    void setHeaders(FileItemHeaders headers);
+
+    default org.apache.commons.fileupload2.core.FileItem toFileUpload2FileItem() {
```
basil, line 225: Bridge from old to new

core/src/main/java/org/apache/commons/fileupload/FileItem.java
```diff
@@ -0,0 +1,438 @@
...
+                    throw e;
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                }
+                return this;
+            }
+        };
+    }
+
+    static FileItem fromFileUpload2FileItem(org.apache.commons.fileupload2.core.FileItem from) {
```
basil, line 332: Bridge from new to old

core/src/main/java/org/apache/commons/fileupload/FileItem.java
```diff
@@ -0,0 +1,438 @@
...
+            }
+
+            @Override
+            public void setHeaders(FileItemHeaders headers) {
+                from.setHeaders(headers.toFileUpload2FileItemHeaders());
+            }
+
+            @Override
+            public org.apache.commons.fileupload2.core.FileItem toFileUpload2FileItem() {
+                return from;
```
basil, line 434: Prevent infinite recursion

core/src/main/java/org/apache/commons/fileupload/FileItemHeaders.java
```diff
@@ -0,0 +1,121 @@
...
+     * Returns an <code>Iterator</code> of all the header names.
+     * </p>
+     *
+     * @return an <code>Iterator</code> containing all of the names of
+     * headers provided with this file item. If the item does not have
+     * any headers return an empty <code>Iterator</code>
+     */
+    Iterator<String> getHeaderNames();
+
+    default org.apache.commons.fileupload2.core.FileItemHeaders toFileUpload2FileItemHeaders() {
```
basil, line 74: Bridge from old to new

core/src/main/java/org/apache/commons/fileupload/FileItemHeaders.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            @Override
+            public Iterator<String> getHeaders(String name) {
+                return FileItemHeaders.this.getHeaders(name);
+            }
+        };
+    }
+
+    static FileItemHeaders fromFileUpload2FileItemHeaders(org.apache.commons.fileupload2.core.FileItemHeaders from) {
```
basil, line 98: Bridge from new to old

core/src/main/java/org/apache/commons/fileupload/FileItemHeaders.java
```diff
@@ -0,0 +1,121 @@
...
+
+            @Override
+            public Iterator<String> getHeaderNames() {
+                return from.getHeaderNames();
+            }
+
+            @Override
+            public org.apache.commons.fileupload2.core.FileItemHeaders toFileUpload2FileItemHeaders() {
+                return from;
+            }
```
basil, line 118: Prevent infinite recursion

core/src/main/java/org/apache/commons/fileupload/InvalidFileNameException.java
```diff
@@ -0,0 +1,61 @@
...
+ * This exception is thrown in case of an invalid file name.
+ * A file name is invalid, if it contains a NUL character.
+ * Attackers might use this to circumvent security checks:
+ * For example, a malicious user might upload a file with the name
+ * "foo.exe\0.png". This file name might pass security checks (i.e.
+ * checks for the extension ".png"), while, depending on the underlying
+ * C library, it might create a file named "foo.exe", as the NUL
+ * character is the string terminator in C.
+ */
+public class InvalidFileNameException extends RuntimeException {
```
basil, line 29: Part of the public API and therefore included for compatibility purposes

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -1056,15 +1062,15 @@ private void parseMultipartFormData() throws ServletException {
...
+        JavaxServletFileUpload<DiskFileItem, DiskFileItemFactory> upload;
         File tmpDir;
         try {
             tmpDir = Files.createTempDirectory("jenkins-stapler-uploads").toFile();
         } catch (IOException e) {
             throw new ServletException("Error creating temporary directory", e);
         }
         tmpDir.deleteOnExit();
-        upload = new ServletFileUpload(new DiskFileItemFactory(DiskFileItemFactory.DEFAULT_SIZE_THRESHOLD, tmpDir));
+        upload = new JavaxServletDiskFileUpload(DiskFileItemFactory.builder().setFile(tmpDir).get());
```
basil, line 1073: The new builder API allows us to omit the default size threshold without any change in behavior. I verified the builder's default is the same number as the old default that we were explicitly passing in.

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -1158,6 +1166,7 @@ private boolean isMultipart() {
     }
 
     @Override
+    @WithBridgeMethods(value = org.apache.commons.fileupload.FileItem.class, adapterMethod = "fromFileUpload2FileItem")
```
basil, line 1169: The Java compiler will generate a `getFileItem(String)` method that returns the new FileUpload 2.x type. The bridge method injector will call the abovementioned method, then call `fromFileUpload2FileItem` to generate the old type. Thus compatibility is retained.

core/src/main/java/org/kohsuke/stapler/Stapler.java
```diff
@@ -1152,6 +1152,20 @@ public FileItem convert(Class type, Object value) {
...
+                if (value == null) {
+                    return null;
+                }
+                try {
+                    return org.apache.commons.fileupload.FileItem.fromFileUpload2FileItem(Stapler.getCurrentRequest().getFileItem(value.toString()));
+                } catch (ServletException | IOException e) {
+                    throw new ConversionException(e);
+                }
+            }
+        }, org.apache.commons.fileupload.FileItem.class);
```
basil, line 1167: I honestly can't remember why I had to do this, but some test failed somewhere along the line back in August 2023 and I added this to fix it. If someone really wants to know the justification, I can try removing this and re-running hours worth of test suites to remember the original motivation, but I'm hoping nobody cares enough to ask this question.

core/src/main/java/org/kohsuke/stapler/StaplerRequest.java
```diff
@@ -497,6 +498,7 @@ public interface StaplerRequest extends HttpServletRequest {
      *      This includes the case where the name corresponds to a simple
      *      form field (like textbox, checkbox, etc.) 
      */
+    @WithBridgeMethods(org.apache.commons.fileupload.FileItem.class)
```
basil, line 501: Bridge methods need to be added to both interface and implementation. See the unit tests for `bridge-method-injector`.

core/src/main/java/org/kohsuke/stapler/Stapler.java
daniel-beck: If I were to be interested in the answer, what test suites are you referring to? Core and/or BOM PCT?

basil: Yes, it was some PCT or ATH test if I remember correctly.

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -1158,6 +1166,7 @@ private boolean isMultipart() {
...
     public FileItem getFileItem(String name) throws ServletException, IOException {
```
BobDu, line 1170: Maybe need deprecated this method and add a new method `getFileItem2` to return a `org.apache.commons.fileupload2.core.FileItem` Object.

Like we do for `FilePath.copyFrom`
https://github.com/jenkinsci/jenkins/pull/9259/files#diff-c1b889a6e9863d548250382cfc6e459ff2ace03c02640ec41c337026384045d8R1187

Otherwise, how can call this method when the plugin has been upgraded to use the fileupload2 api?

If I understand something wrong here, please tell me.

basil: > If I understand something wrong here, please tell me.

You are understanding something wrong here.

PR #557, opened 2024-06-12
Delegate compression to servlet container

### LINKED ISSUE #130: remove gzip compression from stapler

### LINKED ISSUE #265: Replace jzlib with maintained library
jzlib library seems to have been abandoned, so we should not compound the problem by using a fork.

IIRC @jtnord suggested dispensing with the `org.kohsuke.stapler.compression` package. I see it used in e.g. https://github.com/jenkinsci/jenkins/blob/bd7c823c2781d9282eab71a5e2967d386c629344/core/src/main/java/hudson/model/Api.java#L194-L195 and https://github.com/jenkinsci/jenkins/blob/f6378676b11288cc2525852b0a1a441d147e31d3/core/src/main/java/hudson/model/LargeText.java#L214-L219 but I think it is also used for general-purpose page rendering? We could use https://github.com/eclipse/jetty.project/tree/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip instead. Or we could switch to another gzip library, or another encoding altogether. https://en.wikipedia.org/wiki/HTTP_compression#Content-Encoding_tokens suggests that we could use `deflate` with Java Platform calls alone, or Brotli with whatever library. Or we could just remove compression under the assumption that most clients are collocated with the Jenkins service on a fast network.

_Originally posted by @jglick in https://github.com/stapler/stapler/issues/263#issuecomment-899771975_

basil:
Part of a series of three PRs:

- https://github.com/jenkinsci/jenkins/pull/9379
- https://github.com/jenkinsci/stapler/pull/557
- https://github.com/jenkinsci/winstone/pull/390

The primary description of this change is written in https://github.com/jenkinsci/jenkins/pull/9379.

### Testing done

See https://github.com/jenkinsci/jenkins/pull/9379.

Fixes #130
Fixes #265


core/src/main/java/org/kohsuke/stapler/UncaughtExceptionFilter.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+        return h;
+    }
+
+    /**
+     * Disables the effect of {@link #setUncaughtExceptionHandler}, letting all errors be rethrown.
+     */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Legacy switch.")
+    public static boolean DISABLED = Boolean.getBoolean(UncaughtExceptionFilter.class.getName() + ".disabled");
+}
```
basil, line 62: The GitHub diff page makes this unreviewable, as it is a rename of `CompressionFilter` with some simplifications. Here is the actual diff. Please only comment on the changed portions, not the original code which I am not changing.

```java
--- core/src/main/java/org/kohsuke/stapler/compression/CompressionFilter.java	2024-06-12 12:42:12.076182060 -0700
+++ /tmp/UncaughtExceptionFilter.java	2024-06-12 12:42:11.216183675 -0700
@@ -1,4 +1,4 @@
-package org.kohsuke.stapler.compression;
+package org.kohsuke.stapler;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.IOException;
@@ -12,28 +12,7 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-/**
- * Pimps up {@link HttpServletResponse} so that it understands "Content-Encoding: gzip" and compress the response.
- *
- * <p>
- * When exceptions are processed within web applications, different unrelated parts of the webapp can end up calling
- * {@link HttpServletResponse#getOutputStream()}. This fundamentally doesn't work with the notion that the application
- * needs to process "Content-Encoding:gzip" on its own. Such app has to maintain a GZIP output stream on its own,
- * since {@link HttpServletResponse} doesn't know that its output is written through a compressed stream.
- *
- * <p>
- * Another place this break-down can be seen is when a servlet throws an exception that the container handles.
- * It tries to render an error page, but of course it wouldn't know that "Content-Encoding:gzip" is set, so it
- * will fail to write in the correct format.
- *
- * <p>
- * The only way to properly fix this is to make {@link HttpServletResponse} smart enough that it returns
- * the compression-transparent stream from {@link HttpServletResponse#getOutputStream()} (and it would also
- * have to process the exception thrown from the app.) This filter does exactly that.
- *
- * @author Kohsuke Kawaguchi
- */
-public class CompressionFilter implements Filter {
+public class UncaughtExceptionFilter implements Filter {
     private ServletContext context;
 
     @Override
@@ -42,34 +21,18 @@
     }
 
     @Override
-    public void doFilter(ServletRequest _req, ServletResponse _rsp, FilterChain filterChain)
+    public void doFilter(ServletRequest req, ServletResponse rsp, FilterChain filterChain)
             throws IOException, ServletException {
-        Object old1 = swapAttribute(_req, CompressionFilter.class, true);
-
-        CompressionServletResponse rsp = new CompressionServletResponse((HttpServletResponse) _rsp);
-        Object old2 = swapAttribute(_req, CompressionServletResponse.class, rsp);
-
         try {
-            filterChain.doFilter(_req, rsp);
+            filterChain.doFilter(req, rsp);
         } catch (IOException | Error | RuntimeException | ServletException e) {
             if (DISABLED) {
                 throw e;
             }
-            reportException(e, (HttpServletRequest) _req, rsp);
-        } finally {
-            rsp.close();
-
-            _req.setAttribute(CompressionFilter.class.getName(), old1);
-            _req.setAttribute(CompressionServletResponse.class.getName(), old2);
+            reportException(e, (HttpServletRequest) req, (HttpServletResponse) rsp);
         }
     }
 
-    private Object swapAttribute(ServletRequest req, Class<?> key, Object value) {
-        Object old = req.getAttribute(key.getName());
-        req.setAttribute(key.getName(), value);
-        return old;
-    }
-
     private void reportException(Throwable e, HttpServletRequest req, HttpServletResponse rsp)
             throws IOException, ServletException {
         getUncaughtExceptionHandler(context).reportException(e, context, req, rsp);
@@ -92,33 +55,8 @@
     }
 
     /**
-     * Is this request already wrapped into {@link CompressionFilter}?
-     */
-    public static boolean has(ServletRequest req) {
-        return req.getAttribute(CompressionServletResponse.class.getName()) != null;
-    }
-
-    /**
-     * Is this request already wrapped into {@link CompressionFilter},
-     * activate that, so that {@link ServletResponse#getOutputStream()} will return
-     * a stream that automatically handles compression.
-     */
-    public static boolean activate(ServletRequest req) throws IOException {
-        CompressionServletResponse rsp =
-                (CompressionServletResponse) req.getAttribute(CompressionServletResponse.class.getName());
-        if (rsp != null) {
-            rsp.activate();
-            return true;
-        } else {
-            return false;
-        }
-    }
-
-    /**
      * Disables the effect of {@link #setUncaughtExceptionHandler}, letting all errors be rethrown.
-     * Despite its name, this flag does <strong>not</strong> disable {@link CompressionFilter} itself!
-     * Rather use {@code DefaultScriptInvoker.COMPRESS_BY_DEFAULT}.
      */
     @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Legacy switch.")
-    public static boolean DISABLED = Boolean.getBoolean(CompressionFilter.class.getName() + ".disabled");
+    public static boolean DISABLED = Boolean.getBoolean(UncaughtExceptionFilter.class.getName() + ".disabled");
 }
```

core/src/main/java/org/kohsuke/stapler/UncaughtExceptionFilter.java
```diff
@@ -0,0 +1,62 @@
...
+            h = UncaughtExceptionHandler.DEFAULT;
+        }
+        return h;
+    }
+
+    /**
+     * Disables the effect of {@link #setUncaughtExceptionHandler}, letting all errors be rethrown.
+     */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Legacy switch.")
+    public static boolean DISABLED = Boolean.getBoolean(UncaughtExceptionFilter.class.getName() + ".disabled");
```
daniel-beck, line 61: Debatably changed code: Since it's not technically a _legacy switch_ after the package/class name change, justification could be updated (the implementation appears to support modification via script console, which may be considered useful), or the warning addressed by making it `final`.

daniel-beck (approved):

core/src/main/java/org/kohsuke/stapler/StaplerResponse.java
```diff
@@ -229,13 +229,18 @@ default void serveExposedBean(StaplerRequest req, Object exposedBean, ExportConf
      *
      * @param req
      *      Used to determine whether the client supports compression
+     * @deprecated use {@link #getOutputStream}
```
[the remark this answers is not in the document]
basil: Thank you; I have removed the outdated content in commit 82dc11783.

core/src/main/java/org/kohsuke/stapler/UncaughtExceptionFilter.java
basil: Indeed; I have updated the justification in commit ef85d59fc.

MarkEWaite (approved):

PR #571, opened 2024-07-25
Define `HttpResponses.wrap` and use from `RequestImpl.invokeConstructor`

jglick:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/907#discussion_r1691371292.


amuniz (approved):
Verified that it works as expected in https://github.com/jenkinsci/workflow-cps-plugin/pull/907

timja (approved):

PR #575, opened 2024-08-22
Improve debugging support for annotation processors

basil:
While recently attempting to debug an annotation processor, I found that all notices were being discarded, apparently due to an [IntelliJ issue](http://youtrack.jetbrains.net/issue/IDEA-71822) from 13 years ago that appears to have been resolved since then. This PR re-enables logging of notices and adapts the tests to expect them.

### Testing done

`mvn clean verify`

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/jsr269/QueryParameterAnnotationProcessor.java
+++ b/core/src/main/java/org/kohsuke/stapler/jsr269/QueryParameterAnnotationProcessor.java
@@ -69,9 +69,9 @@ public class QueryParameterAnnotationProcessor extends AbstractProcessorImpl {
         }
 
         TypeElement t = (TypeElement) m.getEnclosingElement();
-        FileObject f = createResource(
-                t.getQualifiedName().toString().replace('.', '/') + "/" + m.getSimpleName() + ".stapler");
-        notice("Generating " + f, m);
+        String name = t.getQualifiedName().toString().replace('.', '/') + "/" + m.getSimpleName() + ".stapler";
+        FileObject f = createResource(name);
+        notice("Generating " + name, m);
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/kohsuke/stapler/jsr269/QueryParameterAnnotationProcessor.java
```diff
@@ -69,9 +69,9 @@ private void write(ExecutableElement m) throws IOException {
         }
 
         TypeElement t = (TypeElement) m.getEnclosingElement();
-        FileObject f = createResource(
-                t.getQualifiedName().toString().replace('.', '/') + "/" + m.getSimpleName() + ".stapler");
-        notice("Generating " + f, m);
+        String name = t.getQualifiedName().toString().replace('.', '/') + "/" + m.getSimpleName() + ".stapler";
+        FileObject f = createResource(name);
+        notice("Generating " + name, m);
```
basil, line 74: This was logging an object without a `toString()` method, which made it impossible to expect a specific value in the tests because the default `Object#toString` prints a memory address. By printing the name instead, we have a reliable thing to check for in tests, as well as prettier debug output.

MarkEWaite (approved):

daniel-beck:
Core PR diffs are quite noisy now due to what I suspect is a bug in `analysis-model`, I filed [JENKINS-73709](https://issues.jenkins.io/browse/JENKINS-73709).

basil:
I see. That is annoying indeed. We could always turn this logging back down (though the test cleanup I did could be retained), but I think it is preferable to roll forward and deal with any downstream issues to the degree it is practical. Otherwise, the next person to try to debug these annotation processors will be in the same undesirable position of getting no debug output.

PR #577, opened 2024-08-27
[JENKINS-73690] Avoid null Enum to throw an exception

Wadeck:
[JENKINS-73690](https://issues.jenkins.io/browse/JENKINS-73690)

It's expected to be supported as there is an option to make the QueryParameter required. Positive test cases added as well.

### Testing done
Tested using (new) unit tests, no manual tests outside that.

### Submitter checklist
- [n/a] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
+++ b/core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
@@ -11,9 +11,11 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
+import javax.servlet.ServletException;
 import junit.framework.TestCase;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.jvnet.hudson.test.Issue;
 
 /**
  * @author Kohsuke Kawaguchi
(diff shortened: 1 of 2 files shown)


jeromepochat (approved):

Artmorse (approved):

core/src/test/java/org/kohsuke/stapler/DataBindingTest.java
```diff
@@ -430,4 +432,89 @@ public void testDerivedProperty() {
...
+        MockRequest mr = new MockRequest() {
+            @Override
+            public String getContentType() {
+                return "text/html";
+            }
+        };
+        mr.getParameterMap().put("status", null);
+        RequestImpl req = new RequestImpl(new Stapler(), mr, Collections.emptyList(), null);
+        try {
+            new Function.InstanceFunction(getClass().getMethod("doRequireEnum", StaplerRequest.class, Status.class))
```
timja, line 499: Can you use `assertThrows` for cleaner and less code in both these tests?

Wadeck: I followed the pattern used in that class file, for the `testMismatchingTypes` test case.

I usually prefer consistent style. Only occurrence of assertThrows is in Stapler2Test, compared to ~20 other occurrences of try/catch.

In a new project or if I was refactoring, yes I would have used assertThrows :)

timja: Imo you don't need to add harder to read code because existing code is hard to read.

Wadeck: Depend on the perspective. If you look only at the diff, yes I agree, but if you look at the whole file, having different style make it harder. I expect to have more people reading the whole file compared to people reading only this PR/delta.

PR #604, opened 2024-11-04
Remove JRebel reloading support

basil:
This does not appear to be used by Jenkins.


timja (approved):

jglick:
68dd35512a08dda23ffe8870b01ce358f4b05f37 FTR. I think there was some way you could manually inject this module into Jenkins and then use JRebel for not just pure Java logic but also certain Stapler views? #17 implies that at one point it _was_ active by default. I do not think I ever used it so I do not know more, and I doubt it will be missed.

Issue #416, opened 2022-11-04
@QueryParameter trashes the request body


jtnord:
### Jenkins and plugins versions report

<details>
  <summary>QueryParameter is only supposed to look at the HTTP query but looks also in the RequestBody making it unusable for others</summary>

  
  ```text
  Paste the output here
  ```
</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

windows java11

### Reproduction steps

1. create a method as follows from something accessable from stapler as

```
public void doFoo(@QueryParameter(required = false) String payload, StaplerRequest request, StaplerResponse response) throws IOException {
   String body = IOUtils.toString(req.getInputStream(), req.getCharacterEncoding());
   System.out.print(body);
}
```

call the endpoint with curl `curl theEddpoint --data blahblahblah`

### Expected Results

`blahblahblah` is printed

### Actual Results

the string is enmpty

### Anything else?

because the query parameter is optional it is not in the URL.

[`QueryParameter`](https://github.com/jenkinsci/stapler/blob/3aa95fe179cb5534806b05d2f41947791a78b928/core/src/main/java/org/kohsuke/stapler/QueryParameter.java#L70) calls `request.getParameter(name)`  which is eventually [`ServletRequest.getParameter(String)`](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletRequest.html#getParameter-java.lang.String-)

This code is what will then read the body to find an populate the parameters - meaning the reading of the body later will fail (as it is read the available length of it will be zero)

e.g. for Jetty for example ->
https://github.com/eclipse/jetty.project/blob/bff5b3eb8f5547afd8883029555dd46d06ed7a77/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java#L1046-L1050

This is in violation of the contract of the `QueryParameter`

jtnord:
Note: that in the case of Jetty it first checks the content-type and only does it if the `content-type` if form encoding or multipart data.

timja:
Seems to be a limitation of servlet spec, workarounds here:
https://stackoverflow.com/a/31928740

PR #609, opened 2024-11-14
Fix: Prevent query parameter parsing from consuming request body in S…

haresahani:
This pull request addresses an issue where the use of @QueryParameter in Stapler was causing the request body to be prematurely consumed in some servlet containers (see [#416 ](https://github.com/jenkinsci/stapler/issues/416)).

The fix involves modifying QueryParameter.java to manually parse the query string instead of using request.getParameter(). This change ensures that query parameters are obtained without triggering the consumption of the request body, which is especially important for POST requests that need to retain their body content.

Automated Tests: Ran the existing test suite, which verifies the functionality of @QueryParameter.
Manual Tests:
Tested GET and POST requests with query parameters to confirm they are parsed correctly.
Verified that POST requests with a body retain their content and are unaffected by query parameter processing.
Checked that requests without required query parameters trigger the appropriate exceptions.

Submitter Checklist
 -Created the PR from a topic branch, not the main branch.
 -Provided a descriptive title for the pull request.
 -Linked to the relevant issue: [#416 ](https://github.com/jenkinsci/stapler/issues/416).
 -Added new tests to cover the change, ensuring it works as expected.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/QueryParameter.java
+++ b/core/src/main/java/org/kohsuke/stapler/QueryParameter.java
@@ -70,7 +70,7 @@ public @interface QueryParameter {
                 throw new IllegalArgumentException("Parameter name unavailable neither in the code nor in annotation");
             }
 
-            String value = request.getParameter(name);
+            String value = getQueryParameterFromQueryString(request, name);
             if (a.required() && value == null) {
                 throw new ServletException("Required Query parameter " + name + " is missing");
             }
@@ -78,6 +78,20 @@ public @interface QueryParameter {
                 value = null;
             }
             return convert(type, value);
+        }
+
+        private String getQueryParameterFromQueryString(StaplerRequest2 request, String name) {
+            String queryString = request.getQueryString();
+            if (queryString != null) {
+                // Parse the query string manually
+                for (String param : queryString.split("&")) {
+                    String[] keyValue = param.split("=", 2);
+                    if (keyValue.length == 2 && keyValue[0].equals(name)) {
+                        return keyValue[1];
+                    }
+                }
+            }
+            return null;
         }
     }
 }


core/src/main/java/org/kohsuke/stapler/QueryParameter.java
```diff
@@ -79,5 +79,19 @@ public Object parse(StaplerRequest2 request, QueryParameter a, Class type, Strin
             }
             return convert(type, value);
         }
+
+        private String getQueryParameterFromQueryString(StaplerRequest2 request, String name) {
+            String queryString = request.getQueryString();
```
timja, line 84: I don't think this is right, `QueryParameter` is used to get body parameters as well.

Since we moved from `GET` to `POST` for validation and are moving autocompletion away as well this would break it

timja (changes requested):

core/src/main/java/org/kohsuke/stapler/QueryParameter.java
timja: I see that @jtnord said it was a violation of the contract but Jenkins depends on this behaviour so this change can't be made without a lot of work I would say

jtnord: >  so this change can't be made without a lot of work I would say

Should be limited to to a couple of javascript libraries that handle this?

if we are never going to change the behaviour the least we can do is to fix the declared contract to march reality and tell consumers to not use this when they also want to process the request body.

I have zero recollection of what was the trigger for me to find and file #416

timja: As of https://www.jenkins.io/changelog/2.285/ (2021)

jtnord: Plugins (generally) do not construct the validation, it's part of form binding taglibs from core?

timja: On the server side they ask for parameters in their methods

jtnord: Sure, my point was if the taglibs sent a HTTP POST with query parameters (application/x-www-form-urlencoded) (in addition or replacement of the multipart/form-data rather then this could be fixed.

Now, having thought about this some more, some of the query parameters could be new secrets which would then be exposed by HTTP access logs which would be bad, so following that suggestion would not be a "good thing".™️

haresahani:
Thank you for the detailed feedback and insights! I understand now that the behavior of @QueryParameter—accessing both query and body parameters—is a critical part of Jenkins core and plugin interactions. Based on your comments, I’d like to propose the following steps to address the concerns:

1. Preserving Current Behavior:
Since Jenkins relies on this behavior, I agree that it cannot be altered without significant work and potential breakages. Instead, we can update the documentation or contract for @QueryParameter to reflect this dual behavior clearly. This would inform consumers of the API about the implications when using @QueryParameter with POST requests.

2. Alternative Approach for #416:
If the root issue of #416 (request body consumption) remains valid, we could explore alternate solutions, such as:
Introducing a new annotation (e.g., @QueryOnlyParameter) specifically designed to fetch query parameters without consuming the body.
Providing guidance for consumers on handling scenarios where both query and body parameters are needed.

3. Validation and Secrets Handling:
I acknowledge the potential security concerns raised by @jtnord regarding exposing secrets in query parameters through HTTP logs. This is a valid concern, and any proposed changes should carefully account for such risks.

4. Path Forward:
Given the dependency on the existing behavior, would you agree to:
Update the documentation for @QueryParameter to reflect its dual behavior?
Open a new discussion or issue to explore safer and more robust alternatives to fetch query parameters without consuming the request body?

Please let me know your thoughts, and I’d be happy to adjust the PR or contribute further in line with your guidance. Thank you again for taking the time to review and discuss this in detail.

timja:
From my point of view, option 4 updating the docs makes sense, 2 would be possible if there is a need for it, but I'm not sure there is a need.

PR #612, opened 2024-11-26
Emit Jelly file inside <script> tags

daniel-beck:
PoC quality: This currently makes empty-bodied `<script>` tags into CSP violations.

WDYT? Worth pursuing?

### Testing done

> Screenshot 2024-11-27 at 00 00 22

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did


Wadeck:
FTR from Daniel in Gitter:
>  It's a bit (very) limited given we're restricted to just 40 chars in Firefox at least, but perhaps this might end up useful?

daniel-beck:
> nit suggestion

Seems less safe to do to me.

Also FTR I will need to confirm it's fine to not have `//<![CDATA[` at the start of a block like this.

PR #617, opened 2024-12-19
EE 9 to EE 10

basil:


basil:
These obscure corners of the API are extremely unlikely to be used in practice, and implementing them would be a lot of work—certainly more work than the easier (and better long-term option) path of migrating from EE 8 to EE 9/10. A similar strategy was used for the Acegi migration, which left some obscure corners of the API unimplemented under the assumption that nobody would ever need compatibility for them.

olamy (approved):
LGTM. 
To avoid copy paste of version ,what about using jakartaee bom for jakarta dependencies?

https://repo.maven.apache.org/maven2/jakarta/platform/jakarta.jakartaee-bom/10.0.0/

alecharp:
> These obscure corners of the API are extremely unlikely to be used in practice, and implementing them would be a lot of work—certainly more work than the easier (and better long-term option) path of migrating from EE 8 to EE 9/10. A similar strategy was used for the Acegi migration, which left some obscure corners of the API unimplemented under the assumption that nobody would ever need compatibility for them.

Thanks. I just wanted to make sure I had the correct picture.

alecharp (approved):

basil:
> To avoid copy paste of version ,what about using jakartaee bom for jakarta dependencies?

Consuming that in `org.jenkins-ci.main:jenkins-bom` instead of hard-coding specific versions would eliminate some copypasta, but it also has the potential to introduce problems in consumes of `org.jenkins-ci.main:jenkins-bom` that use things other than `jakarta.servlet-api` and `jakarta.servlet.jsp.jstl-api`. For example, `jakarta.mail-api` is managed in `jakarta.jakartaee-bom`, but we do not want core to specify a particular version of that since it is provided by a Jenkins library plugin. In short, `jakarta.jakartaee-bom` contains too much for us to start shipping it in `org.jenkins-ci.main:jenkins-bom` risk-free.

PR #631, opened 2025-01-24
[JENKINS-70756] Incorrect `RequestImpl.getParameterMap`

jglick:
Noticed in CloudBees CI a stack trace under certain conditions (Move/Copy/Promote) now reproduced in a test

```
java.lang.UnsupportedOperationException
	at java.base/java.util.Collections$UnmodifiableMap.putAll(Collections.java:1665)
	at org.kohsuke.stapler.RequestImpl.getParameterMap(RequestImpl.java:266)
	at org.kohsuke.stapler.RequestImplTest.test_multipart_formdata(RequestImplTest.java:177)
```

First of all, `getParameterMap` is documented to return an immutable map. This was true in `javax.servlet` days, so not obviously a `jakarta.servlet` regression, though of course a particular version of Jetty might have started to enforce immutability.

Secondly, the return type was wrong! The value type is supposed to be `String[]`, not `String`. We were using rawtypes and so did not notice.

Both mistakes seem to date to #53, but _also_ were in the `MockRequest` dating all the way to cfb02d4c4e9a5ae4a68445f19aa7f2253439f57a.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -259,17 +259,29 @@ public class RequestImpl extends HttpServletRequestWrapper implements StaplerReq
     }
 
     @Override
-    public Map getParameterMap() {
-        Map parameterMap = super.getParameterMap();
+    public Map<String, String[]> getParameterMap() {
+        var parameterMap = super.getParameterMap();
         if (isMultipart()) {
             Map<String, String> data = getFormDataFormFields();
-            parameterMap.putAll(data);
+            parameterMap = new HashMap<>(parameterMap);
+            for (var e : data.entrySet()) {
+                var values = parameterMap.get(e.getKey());
+                if (values == null) {
+                    values = new String[] {e.getValue()};
+                } else {
+                    int len = values.length;
+                    var moreValues = Arrays.copyOf(values, len + 1);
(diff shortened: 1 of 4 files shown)


timja (approved):

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
```diff
@@ -259,17 +259,29 @@ public String getParameter(String name) {
     }
 
     @Override
-    public Map getParameterMap() {
-        Map parameterMap = super.getParameterMap();
+    public Map<String, String[]> getParameterMap() {
```
jtnord, line 262: as this is itself implementing the servlet spec, should it not also return an immutable map?

Possibly too risky for a backport, but moving forward for a weekly?

jglick:
Jetty seems to be returning an unmodifiable map dating back to https://github.com/jetty/jetty.project/commit/da627b843fe81fa0fe52a046c1be8595630e9ae7 so not sure how this code could ever have worked.

core/src/main/java/org/kohsuke/stapler/RequestImpl.java
jglick: I think it should indeed.

basil:
Might fix https://github.com/jenkinsci/generic-webhook-trigger-plugin/issues/350.

Issue #483, opened 2023-08-03
Drop support for `side-files`


jglick:
While waiting a long time for a config page to load in a newly started controller, I decided to take a thread dump, and noticed this:

```
"Handling GET /mc/manage/configureClouds/ from … : Jetty (winstone)-… KubernetesCloud/DescriptorImpl/config.jelly PodTemplate/config.jelly" Id=… Group=main RUNNABLE
	at java.base@11.0.20/sun.nio.fs.UnixNativeDispatcher.lstat(UnixNativeDispatcher.java:334)
	at java.base@11.0.20/sun.nio.fs.UnixFileAttributes.get(UnixFileAttributes.java:72)
	at java.base@11.0.20/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:52)
	at java.base@11.0.20/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
	at java.base@11.0.20/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
	at java.base@11.0.20/java.nio.file.Files.readAttributes(Files.java:1764)
	at java.base@11.0.20/java.nio.file.Files.isSymbolicLink(Files.java:2190)
	at org.eclipse.jetty.util.resource.PathResource.checkAliasPath(PathResource.java:98)
	at org.eclipse.jetty.util.resource.PathResource.<init>(PathResource.java:262)
	at org.eclipse.jetty.util.resource.PathResource.addPath(PathResource.java:367)
	at org.eclipse.jetty.server.handler.ContextHandler.getResource(ContextHandler.java:1947)
	at org.eclipse.jetty.webapp.WebAppContext.getResource(WebAppContext.java:410)
	at org.eclipse.jetty.webapp.WebAppContext$Context.getResource(WebAppContext.java:1444)
	at org.kohsuke.stapler.IndexHtmlDispatcher.make(IndexHtmlDispatcher.java:44)
	at org.kohsuke.stapler.MetaClass.buildDispatchers(MetaClass.java:153)
	at org.kohsuke.stapler.MetaClass.<init>(MetaClass.java:99)
	at org.kohsuke.stapler.WebApp$1.computeValue(WebApp.java:230)
	at org.kohsuke.stapler.WebApp$1.computeValue(WebApp.java:227)
[...]

PR #637, opened 2025-02-03
Remove Support for side-files and Update maven-example

ahmed-kamal2004:
:wave:
 - According to issue #483,` side-files` feature is not used in Jenkins, and it results in wasting a part of time crawling the filesystem searching for `index.html` side-files

Screenshot from 2025-01-30 01-36-18

So I removed that snippet of code
`Dispatcher d = IndexHtmlDispatcher.make(webApp.context, clazz);`
`        if (d!=null)`
`            dispatchers.add(d);`

this removes support for `web-inf/side-files/index.html` pages

---
- update maven-example

update jetty contextpath to `/` and remove side-files as they no longer supported beginning from this pull request.

example when sending request to local-instance of the example at `localhost:9090/items/`
Screenshot from 2025-01-31 16-52-18

---
- use `discover` instead of 'discoverExtensions' following the `TODO` in the code.

---
- Note: afaik jenkins now is using only (`html`-`jelly`-`groovy`) files, there is no `jsp` files
so we can't remove the `facets` support
Screenshot from 2025-01-31 17-20-29
Screenshot from 2025-01-31 17-21-41
Screenshot from 2025-01-31 17-22-08
Screenshot from 2025-01-31 17-22-37

some snippets showing the used extensions in jenkins, and there is no `.jsp`

> This pull request is a new one, I removed the previous for a better commit history.

pls review if this pull request could be merged @timja @jglick :100:

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Facet.java
+++ b/core/src/main/java/org/kohsuke/stapler/Facet.java
@@ -73,7 +73,7 @@ public abstract class Facet {
     /**
      * Discovers all the facets in the classloader.
      */
-    public static List<Facet> discover(ClassLoader cl) {
+    public static List<Facet> discover(ClassLoader... cl) {
         return discoverExtensions(Facet.class, cl);
     }
 
--- a/core/src/main/java/org/kohsuke/stapler/WebApp.java
+++ b/core/src/main/java/org/kohsuke/stapler/WebApp.java
@@ -183,10 +183,8 @@ public class WebApp {
         this.servletContext = context;
         this.context = context != null ? ServletContextWrapper.fromJakartServletContext(context) : null;
         // TODO: allow classloader to be given?
-        facets.addAll(Facet.discoverExtensions(
-                Facet.class,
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/kohsuke/stapler/MetaClass.java
```diff
@@ -156,11 +156,6 @@ public class MetaClass extends TearOffSupport {
             f.buildIndexDispatchers(this, dispatchers);
         }
 
-        Dispatcher d = IndexHtmlDispatcher.make(webApp.getServletContext(), clazz);
```
jglick, line 159: But you did not delete `IndexHtmlDispatcher` nor `StaticViewFacet`.

jglick:
> there is no `jsp` files
> so we can't remove the `facets` support

Sorry, I did not follow what this was referring to. The `stapler-jsp` module? That would make sense to delete, since it is unused AFAIK, but is unrelated to `side-files` so far as I can tell.

jglick:
> within this month

Sorry, I will not be available in the near future.

PR #655, opened 2025-04-03
Escape values inside JS string

daniel-beck:
Someone reported to us that the URL portion was not properly escaped in URLs like the following, resulting in broken JS:

`http://localhost:8080/$stapler/bound/script/foo'bar'baz?var=proxy&methods=foo,bar`

Previously output was:
`proxy = makeStaplerProxy('/foo'bar'baz','a534b6b86f174fb6a2180a8cc9ad94589eabfc75149c2e1faec6fc2a5448a275',['bar','foo']);`

Now output is:
`proxy = makeStaplerProxy('/foo\'bar\'baz','a534b6b86f174fb6a2180a8cc9ad94589eabfc75149c2e1faec6fc2a5448a275',['bar','foo']);`

This is in the code branch for `WithWellKnownURL` which hasn't had an implementation in about a decade as far as I'm aware, and even before it wasn't in a relevant plugin. This fixes the case for when there's ever a `WithWellKnownURL` that has single quotes in its URL.

There is no security impact, so this is merely a minor bug fix.

Amends https://github.com/jenkinsci/stapler/pull/385.

### Testing done

Accessing the above URLs.

Core PR with test coverage: https://github.com/jenkinsci/jenkins/pull/10500

### Submitter checklist

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/bind/Bound.java
+++ b/core/src/main/java/org/kohsuke/stapler/bind/Bound.java
@@ -144,8 +144,8 @@ public abstract class Bound implements HttpResponse {
         return "makeStaplerProxy('" + escapeQuotedString(url) + "','" + crumb + "',[" + methodNamesList + "])";
     }
 
-    private static String escapeQuotedString(String singleQuotedJsValue) {
-        return singleQuotedJsValue.replace("'", "\\'");
+    static String escapeQuotedString(String singleQuotedJsValue) {
+        return singleQuotedJsValue.replace("\\", "\\\\").replace("'", "\\'");
     }
 
     private static String camelize(String name) {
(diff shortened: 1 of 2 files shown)


timja (approved):

Kevin-CB (changes requested):
Be sure to also escape `\` otherwise you can bypass this restriction by adding another `\` before the simple quote.

`http://localhost:8080/$stapler/bound/script/foo'bar%5C'baz?var=proxy&methods=foo,bar`

==>

`proxy = makeStaplerProxy('/foo\'bar\\'baz','8d54536ecc972bed5710a2ef04afaf727e43261edff68f22af14f4d370a3313c',['bar','foo']);`

daniel-beck:
> Be sure to also escape `\` otherwise you can bypass this restriction by adding another `\` before the simple quote.

Good point. After https://github.com/jenkinsci/jenkins/pull/10344 it's possible to have `%5c` again without HTTP 400 Suspicious Path Character.

Kevin-CB (approved):
Restested, it works as expected.

NotMyFault (approved):

timja:
🚢 @daniel-beck ?

daniel-beck:
@timja I like not overloading a single (Jenkins) release with multiple updates to the same component, and neighboring PRs got in earlier whenever I checked.

PR #660, opened 2025-04-22
[JENKINS-75468] Omit null elements from collections during databinding

rrjjvv:
Elements that convert to `null` are considered invalid/unwanted, so adding those `null` values is invalid/unwanted as well.

This fixes the the underlying issue from https://issues.jenkins.io/browse/JENKINS-75468.

The initial attempt to fix that issue (https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/778) led to discussions indicating it was an issue with stapler and not individual plugins.

Rather than doing the filtering at every point where elements were added to a collection, I opted to effectively create a collection that did not allow `null` elements.  A simpler solution would be to modify `Lister` itself, but it (technically) lives outside the Jenkins umbrella.

### Testing done
There is a new unit test which approximates the original issue.  This new test passes, all existing tests pass, and core Jenkins tests continue to pass when using this new version.  The change will impact four various ways of "add something to a list", but only explicitly tests one of them (the one relevant to the bug being fixed).  Testing for every path would take more effort and might be seen as over-specification, so I opted for simplicity until reviewed.

I originally used `DataBoundResolvable` to mimic Jenkins-specific `Descriptor#newInstance` (since both leave the decision to return `null` to the class itself, and both document `null` as a valid return), but eventually opted to just use `BindInterceptor` since that was simpler and matched surrounding code.

As well, a manual test for the desired behavior in the ticket works (using this new stapler in lieu of the modified plugin.

### Submitter checklist

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -814,7 +814,9 @@ public class RequestImpl extends HttpServletRequestWrapper implements StaplerReq
                 return a;
             }
 
-            Lister l = Lister.create(type, genericType);
+            // It is legitimate for a JSON object to convert to null (via BindInterceptor, DataBoundResolvable, etc.),
+            // but that is intended to signify an invalid/undesired object.  Being invalid, omit them from collections.
+            Lister l = createNullFreeLister(type, genericType);
 
             if (o instanceof JSONObject j) {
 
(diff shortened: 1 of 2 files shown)


rrjjvv:
Also, the formatting of that test isn't great, but spotless rejected every attempt I made to make it nicer.

jglick (approved):
Looks right, and `bom` passes with this.

PR #663, opened 2025-05-06
[JENKINS-16341] Automatically expire entries from `BoundObjectTable`

jglick:


core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -260,38 +254,22 @@ public Object getDynamic(String id) {
             return resolve(id);
         }
 
-        private synchronized Ref release(String id) {
-            return entries.remove(id);
+        /**
+         * Releases a particular id from the table.
+         * Normally would be called via {@link #releaseMe} which infers the id from the current request.
+         */
+        public void release(String id) {
```
jglick, line 261: Making this public so we can clean up the reflection in https://github.com/jenkinsci/jenkins/commit/aac8c239721ed52caacbff389a29d5db2c1e3a00.

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -150,22 +147,11 @@ public Table getStaplerFallback() {
         return resolve(false);
     }
 
-    private Bound bind(Ref ref) {
```
jglick, line 153: inlined for clarity

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -150,22 +147,11 @@ public Table getStaplerFallback() {
...
      * Binds an object temporarily and returns its URL.
      */
     public Bound bind(Object o) {
-        return bind(new StrongRef(o));
-    }
-
-    /**
-     * Binds an object temporarily and returns its URL.
-     */
-    public Bound bindWeak(Object o) {
```
jglick, line 167: Apparently never used.

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -213,7 +199,6 @@ public Table getTable() {
      */
     public static class Table implements Serializable {
         private final Map<String, Ref> entries = new HashMap<>();
-        private boolean logging;
```
jglick, line 216: Unnecessary, just use regular tools like https://plugins.jenkins.io/log-cli/

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -226,11 +211,20 @@ private synchronized Bound add(Ref ref) {
...
+            // Before adding a new entry, clean up any old ones:
+            var it = entries.entrySet().iterator();
+            while (it.hasNext()) {
+                var entry = it.next();
+                if (!entry.getValue().valid()) {
+                    LOGGER.fine(() -> "removing stale " + entry.getKey() + ": "
+                            + entry.getValue().get());
+                    it.remove();
+                }
+            }
```
jglick, line 223: The fix.

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -333,54 +311,48 @@ public void generateResponse(StaplerRequest2 req, StaplerResponse2 rsp, Object n
...
-
     /**
-     * True to activate debug logging of session fragments.
+     * How long (in milliseconds) a {@link StrongRef} remains valid for creation if it is not retrieved.
      */
-    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Legacy switch.")
-    public static boolean DEBUG_LOGGING = Boolean.getBoolean(BoundObjectTable.class.getName() + ".debugLog");
+    private static final long EXPIRATION_TIME = Long.getLong(
+            BoundObjectTable.class.getName() + ".EXPIRATION_TIME",
+            Duration.ofMinutes(15).toMillis());
```
jglick, line 352: Seems long enough for practical purposes. After this point, unused render-on-demand fragments will not work. For example, if you opened a project config form and came back 15m later and selected a parameter type from the dropdown, it will not render the per-parameter config snippet.

timja (approved):

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
daniel-beck: Are we prepared to revert this based on user feedback? Because I'd be surprised if this didn't happen with some regularity.

daniel-beck: WDYT about https://github.com/jenkinsci/stapler/pull/665 + https://github.com/jenkinsci/jenkins/pull/10631?

jglick: If it works, sure. Or we can just increase the timeout to several hours, at which point your session is likely to have expired anyway so this would just be a fallback defense.

daniel-beck: Good point, cleaning up after 24 hrs or so (this approach) should not affect users while my PR keeps memory use low in the mean time, allowing for the greater timeout 👍

daniel-beck (approved):
Looks like a reasonable change to me. The current 1d timeout should be safe enough, and if that still accomplishes the goal, that's great.

PR #665, opened 2025-05-08
Allow releasing bound objects via HTTP

daniel-beck:
Experimental alternative to https://github.com/jenkinsci/stapler/pull/663.

I expect time-based expiration to be problematic (https://github.com/jenkinsci/stapler/pull/663#discussion_r2078958053) so this instead allows releasing bound objects when navigating away from the page that registered the proxies.

Pros: Proxies continue to exist until you leave the page.
Cons: Browser involvement needed, a page can accumulate proxies in badly-written code. Probably more.

The current implementation sends one request per proxy, could be collapsed to just one if this approach is deemed preferable.

### Testing done

1. Script console from https://github.com/jenkinsci/jenkins/pull/10623#discussion_r2075789533: Empty
2. Navigate to a job config screen
3. Script console: Very much not empty
4. Navigate away via breadcrumbs
5. Script console: Empty again

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
+++ b/core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
@@ -76,6 +76,27 @@ public class BoundObjectTable implements StaplerFallback {
     }
 
     /**
+     * Explicitly unbind this object via HTTP.
+     *
+     * @throws HttpResponses.HttpResponseException expected outcome returning 200 OK
+     */
+    public void doRelease(StaplerRequest2 req, StaplerResponse2 rsp) throws HttpResponses.HttpResponseException, IOException {
+        final Table table = resolve(false);
+        if (table == null) {
+            rsp.sendError(404);
+        }
+
+        String id = req.getRestOfPath().replace("/", "");
+
+        Object object = table.resolve(id);
+        if (object == null) {
+            rsp.sendError(200);
+        }
+        table.release(id);
+        rsp.sendError(200);
+    }
+
+    /**
      * This serves the script content for a bound object. Support CSP-compatible st:bind and similar methods of making
      * objects accessible to JS.
      *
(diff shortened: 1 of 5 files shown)


jglick:
Fine it it works.

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
```diff
@@ -75,6 +75,30 @@ public static boolean isValidJavaIdentifier(String name) {
                 .allMatch(it -> Character.isJavaIdentifierPart(it) && !Character.isIdentifierIgnorable(it) && it < 256);
     }
 
+    /**
+     * Explicitly unbind this object via HTTP.
+     *
+     * @throws HttpResponses.HttpResponseException expected outcome returning 200 OK
+     */
+    public void doRelease(StaplerRequest2 req, StaplerResponse2 rsp)
```
jglick, line 83: `@POST`?

jglick:
I wonder if it would not be better to just deprecate `l:renderOnDemand`, and all uses of `BoundObjectTable` other than `ProgressiveRendering` which already has its own logic for releasing the object when the user navigates away from the page (currently done on the server side simply by checking whether a request has been received recently). The usual uses of `l:renderOnDemand` are to lazily render config form snippets, for newly instantiated objects, but these could be done more efficiently by creating an `l:ajax` page tied to the `Descriptor` since its contents are not sensitive to context. (Just need to make sure it is not used in case where there _is_ already an `instance`—normally these are rendered eagerly; and to honor `DescriptorByNameOwner` when constructing URLs, to ensure compatibility with the CloudBees CI Templates plugin.)

core/src/main/java/org/kohsuke/stapler/bind/BoundObjectTable.java
daniel-beck: I guess, although I expect this would be replaced anyway due to

> The current implementation sends one request per proxy, could be collapsed to just one if this approach is deemed preferable.

PR #702, opened 2025-08-10
Fix skipping for File and Gzip sources of LargeText

das7pad:
While working on another patch for https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/704 I noticed subtle bugs in the current implementation of the file-based sources for `LargeText`. Let me know if you want me to create a ticket here for these bugs.

The skipping is used via `?startByte=xxx` in various implementations of log-tailing in Jenkins core/plugins.

Seeking past the end is allowed in a RandomAccessFile. The behavior is desired there for creating files with "holes". A start greater than the current file size will simply result in empty reads. This differs from the buffer source that throws `EOFException` when requesting logs past the end of the file. This may seem negligible at first, but the returned end position after writing the log is also incremented and using the return value as start for following reads would skip over bytes. E.g. Assume a log file with 100 bytes, `writeLogTo(42)` -> copy bytes 42-100 and return 100;  `writeLogTo(1000)` -> copy 0 bytes and return 1000 (!). Following reads would start at 1000 and skip over bytes 100-1000.

Gzip: When reaching the EOF, skip will return -1. The while loop will increment the start at every turn. It might eventually overflow and finish. GZIPInputStream.skip already skips in a loop and we can drop the outer loop in GzipAwareSession.

I've added another commit for aligning all the implementations and check the return value at the call-site (in `writeLogUncounted`). This greatly simplifies things: The signatures of the methods in the `LargeText.Session` interface are now matching those in `InputStream`. Let me know if you prefer a separate PR for this change.

### Testing done

I've extended the tests by:
- using all the source variants of `LargeText` . It is worth noting that the file/Gzip tests are failing/hanging on the main branch when seeking past the end.
- reading a log that requires multiple iterations in `writeLogUncounted` for each of the source variants.
- extending the on-demand/ignored test with a write at the end that is read back. I've also wrapped the test in try/finally for removing the tmp file on error.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


core/src/test/java/org/kohsuke/stapler/framework/io/LargeTextTest.java
```diff
@@ -46,69 +47,137 @@ public class LargeTextTest {
...
+        // Large string with rest after reading multiples of 1024 bytes.
+        String large = "Hello World! ".repeat(1025).trim();
+        assertEquals(large, tail(t, large, 0));
+        assertEquals(large.substring(1337), tail(t, large, 1337));
     }
 
-    String tail(String text, long start) throws IOException {
-        LargeText t;
-        try (ByteBuffer bb = new ByteBuffer()) {
+    static class fromBuffer implements BuildLargeText {
```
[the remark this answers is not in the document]
das7pad: Thanks for the suggestion, done in 57a29b41033cc1c2ee6a44b503142fa2644c37e8.

core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
```diff
@@ -530,9 +532,14 @@ public void close() throws IOException {
...
+        /**
+         * Like {@link RandomAccessFile#skipBytes} but with long instead of int.
+         */
         @Override
-        public void skip(long start) throws IOException {
-            file.seek(file.getFilePointer() + start);
+        public long skip(long start) throws IOException {
+            long to = Math.min(file.length(), file.getFilePointer() + start);
+            file.seek(to);
+            return to;
```
jglick, line 542: This is supposed to be returning the number of bytes skipped, not the resulting offset, IIUC.

das7pad: Good spot, fixed in c964db5681e061f0915d611b804d47d2052fa59c. Thanks!

Side note: It did not affect the behavior. There is only one usage of skip() at the start of streaming. The file position is always 0 in this case.

PR #718, opened 2025-09-14
Add interface based constructor for LargeText

das7pad:
- Part of https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/704

As suggested in https://github.com/jenkinsci/stapler/pull/703#pullrequestreview-3107730846 and based on the guidance in https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/704 to avoid hacks in favor of upgrading the minimum Jenkins version for plugins, this PR is adding an interface based constructor for `LargeText`.

It will be used by the `workflow-api` plugin for a streaming based step log implementation.

(https://github.com/das7pad/jenkins-workflow-api-plugin/commit/71f6322ee1e754e4777f33b0a050fbf0aee6e785#diff-fa50e2c142db03c7b1d36cd1c8bf3be413f1e375ef7d0f28ca0e1889c49238d8R384-R392 -- With this PR merged and stapler upgraded in a Jenkins weekly, we can pull it into the plugin and avoid the hack.)

### Testing done

#702 has added extensive tests for all the different `LargeText` source types. I've added another test for the interface based test.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
+++ b/core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
@@ -63,7 +63,7 @@ public class LargeText {
     /**
      * Represents the data source of this text.
      */
-    private interface Source {
+    public interface Source {
         Session open() throws IOException;
 
         long length();
@@ -167,6 +167,12 @@ public class LargeText {
                 return true;
             }
         };
+        this.completed = completed;
+    }
+
+    public LargeText(Source source, Charset charset, boolean completed) {
+        this.charset = charset;
+        this.source = source;
         this.completed = completed;
     }
 
@@ -509,7 +515,7 @@ public class LargeText {
      * Represents the read session of the {@link Source}.
      * Methods follow the contracts of {@link InputStream}.
      */
-    private interface Session extends Closeable {
+    public interface Session extends Closeable {
         long skip(long n) throws IOException;
 
         int read(byte[] buf) throws IOException;
--- a/core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
+++ b/core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
@@ -510,7 +510,7 @@ public class LargeText {
      * Methods follow the contracts of {@link InputStream}.
      */
     private interface Session extends Closeable {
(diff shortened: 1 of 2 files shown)


timja:
Can you open the PR to workflow api please, you can use the incremental version in Jenkins core:
https://github.com/jenkinsci/stapler/pull/718/checks?check_run_id=50334233013

and then the Jenkins core dependency in workflow API, so we can see it working end 2 end.

core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
```diff
@@ -509,8 +515,8 @@ public boolean isFull() {
      * Represents the read session of the {@link Source}.
      * Methods follow the contracts of {@link InputStream}.
      */
-    private interface Session extends Closeable {
-        long skip(long start) throws IOException;
+    public interface Session extends Closeable {
+        long skip(long n) throws IOException;
```
timja, line 519: can we document these methods now that the interface is public?

core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
```diff
@@ -63,7 +63,7 @@ public class LargeText {
     /**
      * Represents the data source of this text.
      */
-    private interface Source {
+    public interface Source {
         Session open() throws IOException;
```
timja, line 67: can we document these methods now that the interface is public?

core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
das7pad: Done in 50ce95ca28a81a072262f166a08cf85799045f83

core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java
das7pad: Done in 50ce95ca28a81a072262f166a08cf85799045f83

jglick:
@das7pad seems you are doing something in `workflow-api-plugin`; please file that as a draft PR.

PR #769, opened 2026-05-23
Avoid NPE in `getFileItem2` when request is not multipart

linesight:
`RequestImpl.getFileItem2` calls `parseMultipartFormData()` unconditionally. With no `Content-Type` on the request, Commons FileUpload 2.x throws an internal NPE rather than `FileUploadContentTypeException`, so it escapes the catch chain.

Other call sites (`getParameter`, `getParameterMap`, ...) already guard with `isMultipart()`; just lining up `getFileItem2` with them.

Sibling of #651 from the FileUpload 1.x → 2.x migration. Surfaced downstream via [jenkinsci/file-parameters-plugin#199](https://github.com/jenkinsci/file-parameters-plugin/issues/199) on POST `/job/.../buildWithParameters` from a .NET client that doesn't send `Content-Type`

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
+++ b/core/src/main/java/org/kohsuke/stapler/RequestImpl.java
@@ -1340,6 +1340,9 @@ public class RequestImpl extends HttpServletRequestWrapper implements StaplerReq
 
     @Override
     public FileItem getFileItem2(String name) throws ServletException, IOException {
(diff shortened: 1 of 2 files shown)


MarkEWaite (approved):
Thanks!  I added a test that shows the problem when run on the master branch and shows the problem is fixed with this pull request.

PR #786, opened 2026-07-31
Fix Function class-init deadlock under concurrent class loading

apuig:
https://github.com/jenkinsci/stapler/blob/690efb4d25e9af6ef196a243ea7a7054992c56fd/core/src/main/java/org/kohsuke/stapler/Function.java#L229-L231

`Function.<clinit>` constructed a `StaticFunction` (extends `MethodFunction extends Function`), demanding `LC[Function$MethodFunction]` while `Function` was still initialising. Threads entering `ClassDescriptor.<init>` concurrently to build a `Function.InstanceFunction` held `LC[Function$MethodFunction]` and waited for `LC[Function]`, creating a permanent three-thread cycle:

```
Thread A  holds LC[Function],                waiting for LC[Function$MethodFunction]
Thread B  holds LC[Function$MethodFunction], waiting for LC[Function$InstanceFunction]
Thread C                                     waiting for LC[Function]
```

This is a standard class-init cycle per [JVMS 5.5](https://docs.oracle.com/javase/specs/jvms/se25/html/jvms-5.html#jvms-5.5), theoretically possible on any JDK. It was first observed on JDK 25 where [JEP 483](https://openjdk.org/jeps/483) (AOT/CDS, active by default: dump shows `mixed mode, sharing` without any explicit `-Xshare` flag) pre-links inner class metadata before `<clinit>` runs, widening the race window enough to trigger consistently.

Observed in propietary ATH.

## Fix

Move `PARSE_METHODS` and `RETURN_NULL` into a private `ParseMethodsHolder` inner class (lazy-holder idiom). The holder's `static {}` block runs only on first access, after `Function` and all its inner classes are fully initialised. No behaviour change on any JDK.

## Test

No unit test can trigger the race deterministically. Any classloader-based approach serialises on `LC[Function]`: all threads wait for one to finish `Function.<clinit>`, closing the window. The race requires JEP 483 pre-linking inner class metadata independently of `<clinit>`, which is JVM-internal and cannot be forced from user code. 

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Function.java
+++ b/core/src/main/java/org/kohsuke/stapler/Function.java
@@ -202,8 +202,8 @@ public abstract class Function {
                 }
 
                 // if the databinding method is provided, call that
-                Function binder = PARSE_METHODS.get(t);
-                if (binder != RETURN_NULL) {
+                Function binder = ParseMethodsHolder.PARSE_METHODS.get(t);
+                if (binder != ParseMethodsHolder.RETURN_NULL) {
                     arguments[i] = binder.bindAndInvoke(null, req, rsp);
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -218,62 +218,67 @@ Object bindAndInvoke(Object o, StaplerRequest2 req, StaplerResponse2 rsp, Object
         return invoke(req, rsp, o, arguments);
     }
 
-    /**
-     * Computing map that discovers the static 'fromStapler' method from a class.
```
jglick, line 222: (ignore WS)

jglick (approved):
or #787

apuig:
Tested #787 and it properly fixes the issue. Preferring it as the cleaner solution (no extra holder class).

PR #787, opened 2026-07-31
Move `RETURN_NULL` into `StaticFunction` to prevent class loading deadlock

jglick:
_Untested_ alternative to #786 that is arguably a bit simpler.

### DIFF
--- a/core/src/main/java/org/kohsuke/stapler/Function.java
+++ b/core/src/main/java/org/kohsuke/stapler/Function.java
@@ -203,7 +203,7 @@ public abstract class Function {
 
                 // if the databinding method is provided, call that
                 Function binder = PARSE_METHODS.get(t);
-                if (binder != RETURN_NULL) {
+                if (binder != StaticFunction.RETURN_NULL) {
                     arguments[i] = binder.bindAndInvoke(null, req, rsp);
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -222,59 +222,44 @@ Object bindAndInvoke(Object o, StaplerRequest2 req, StaplerResponse2 rsp, Object
      * Computing map that discovers the static 'fromStapler' method from a class.
      * The discovered method will be returned as a Function so that the invocation can do parameter injections.
      */
-    private static final ClassValue<Function> PARSE_METHODS;
+    private static final ClassValue<Function> PARSE_METHODS = new ClassValue<>() {
```
jglick, line 225: This I suppose should be fine.

core/src/main/java/org/kohsuke/stapler/Function.java
```diff
@@ -222,59 +222,44 @@ Object bindAndInvoke(Object o, StaplerRequest2 req, StaplerResponse2 rsp, Object
...
+                default:
+                    throw new IllegalArgumentException("Too many 'fromStapler' methods on " + from);
             }
-        };
-    }
+        }
+    };
 
-    /**
-     * @see StaticFunction#RETURN_NULL
```
jglick, line 276: Javadoc was meaningless anyway—referring to a `private` (undocumented) member.

apuig (approved):
I was not able to reproduce the issue in this version

jglick:
In general, a Java class initializer should never try to load a subtype, or otherwise have a cycle. Unfortunately the language does not detect this statically,and I guess static analysis tools miss it too. Not the first time I have come across this kind of situation. From a quick search:
- https://github.com/jenkinsci/jenkins/pull/8736
- https://github.com/jenkinsci/jenkins/pull/4714
- https://github.com/jenkinsci/jenkins/pull/5932
- https://github.com/jenkinsci/jenkins/pull/6449

